PR #6, opened 2017-10-12
Add echo test in .travis.yml

bjorgve:


.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 
 before_install: 
 
-- apt-get install libeigen3-dev
+- sudo apt-get install libeigen3-dev
```
robertodr: You can't use `sudo` on the `sudo: false` machines on Travis. Try doing this to install dependencies:
```
addons:
  apt:
    packages:
    - cmake
    - time
```

stigrj:
I went for codecov instead of coveralls, simply because I think i looks better. @robertodr, any thoughts on this? The codecov slack notification does not seem to work. Another annoyance is that codecov.yml currently cannot be prepended with a dot, but I trust this will be corrected in the future,  [https://github.com/codecov/support/issues/429](url).

bast:
I think both are fine and if you prefer codecov, that is a good reason to use it. I personally like the "punk" appearance of coveralls but it's a personal taste thing.

robertodr:
Looks good to me :smiley:

.travis.yml
```diff
@@ -1,2 +1,25 @@
 language: cpp
 sudo: false
+
+notifications:
+  slack:
+    secure: 1jnJfgRwfuF8dzfN5daTItQ+jrilhRSjz88gKMhMpqvWrpaAZrytTcejs5FELXadeLIiddoT7c5GKLrCngVzadMi60bc/kfE2Lzer8z31i6mfPXGLvd3zWpVa6Tt8Bur2wL/oB/f6herDQurW2t0HcTp6oc6Mv3iI23PPl/MQG4w6oNKc6SOeBIBDykJOSaf3+SY/z4vXBqqF5lbmWHUIngTUkV7UEzmj+c4MukOUn9IyYjNSgvZd1ql4SKG58gzkafB1tbhxPjyJqLcWjokwlhVYTxHulBqemxRj7DGznWAUL72Pbc6ot27R655wvl/OEoL7X8U0aLIJIlIzLl49WDv5c81wGw/iOVjj5leBpfJl6QlyNMJC0jPEEJYPbEWxzgZGNa1eYbZyroOKvBBalcYVzxq018VgwYjMdZL0zFhdTymgFcDnR1p3Z1Lii6fN3oO2QVAsyBq3sN+IVtlyAWnHIGP+eyEGpggKJ1GsnG/Mb8cjBc6hjG8Q3fRQEfB5C5FeWJSU67hTnuGpTWKcF8rcfbtrw5D1JzBkqYjL6PuG5/VLu8Ro/IihBKf0N+sFa5GvdmbWOEWDEpiIMfWvcOpTIabZyRzcLJX7fqP29b2lUjoLVT3g5+cgSYSZcVyCFj+kunL1LZpgsBf9tr6r+OWzXwNVti+fEvF0KqMHHs=
```
robertodr: @stigrj Did you encrypt the token?

codecov.yml
```diff
@@ -0,0 +1,15 @@
+coverage:
+  notify:
+    slack:
+      default:
+        url: "secret:bInnBtjHWqcMiw+KB/NyVFU2yk6gHLdMLmgIdh1wT/VK0OetNc+WP2gYxhQwEy6yoJyRBSmpMfKwn4RuVSEe6orVQNMpt3NS9g53B2btW7a1Iv3SCl+GhPVkVHQiwarhWgFQnbR4QYQyBI5wTOOC1QanuUVn0yQ05FMz/VnMRU8="
```
robertodr: Same here, is this the encrypted secret token?

.travis.yml
```diff
@@ -1,2 +1,25 @@
...
+addons:
+  apt:
+    packages:
+      - libeigen3-dev
+      - lcov
+
+script:
+  - ./setup --enable-tests --coverage
+  - cd build
+  - make
```
robertodr: For cross-platform compatibility use `cmake --build . -- VERBOSE=1 -j2`. This will call the appropriate build tool for the chosen generator - Unix Makefiles - in this case and pass it the flags after the `--`

.travis.yml
```diff
@@ -1,2 +1,25 @@
...
+  - cd tests
+  - ./unit_tests.x
```
robertodr: I would suggest using `ctest --output-on-failure --verbose` instead of `./unit_tests.x`, once again for cross-platform compatibility. You will need to set up CTest to execute the unit tests, though.

codecov.yml
stigrj: Yes, this one is encrypted using codecov. Not sure if it works, though, as I am yet to receive any notifications. Could be the "threshold: 1%" that prevents notifications.

.travis.yml
stigrj: Yes, encrypted using "travis encrypt"

.travis.yml
stigrj: I will try this out

.travis.yml
stigrj: I had the same impression, but ctest did not work *immediately* out of the box. I will give it another shot.

stigrj:
@robertodr or @bast , what does "addons: &1" and "addons: *1" mean?

robertodr:
In YAML you can have tag sections and refer to them later on. `&1` tags the section, `*1` will reuse that section. It's a convenient way to make the `.travis.yml` more compact, because you can fill up the `addons` section once and reuse it multiple times.

PR #9, opened 2017-11-29
Add simple MPI features

stigrj:
The code can now be compiled with MPI support. The inherent parallelization of MRCPP is still only using shared memory OpenMP, so there is no MPI work/data distribution _inside_ MRCPP. This patch provides an API for doing work/data distribution in the host program by sending complete function objects between MPI processes. The API is very simple: An MPI communicator must be initialized (and finalized) by the host program and assigned to the MRCPP global communicator ``mpi::world``:
```cpp
MPI_Init(&argc, &argv);
mpi::world = MPI_COMM_WORLD; // Assign to MRCPP global communicator
int world_size = mpi::world.size(); // Get total number of ranks
int world_rank = mpi::world.rank(); // Get my rank id
MPI_Finalize();
```
Work distribution is achieved brute force by assigning different functions to particular ranks:
```cpp
FunctionTree<3> tree(MRA); //All ranks define the tree
if (mpi::world.rank() == 0) project(tree, f); //Only rank 0 projects the function
```
A ``FunctionTree`` can be bundled up in a ``mpi::tree_request`` along with a unique tag and sent between ranks (blocking communication):
```cpp
mpi::tree_request<3> request(tree, 111); //Create request to send tree with unique tag  
if (mpi::world.rank() == 0) mpi::world.send_tree(request, 1); //Rank 0 sends to 1
if (mpi::world.rank() == 1) mpi::world.recv_tree(request, 0); //Rank 1 receives from 0 
```
This completes the API.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake/downloade
 include(autocmake_cxx)
 include(GNU.CXX)
 include(Intel.CXX)
+include(autocmake_mpi)
 include(autocmake_omp)
 include(autocmake_safeguards)
 include(autocmake_version)
--- a/cmake/autocmake.yml
+++ b/cmake/autocmake.yml
@@ -16,6 +16,7 @@ modules:
       - 'compiler_flags/Intel.CXX.cmake'
   - plugins:
     - source:
+      - '%(url_root)modules/mpi.cmake'
       - '%(url_root)modules/omp.cmake'
       - '%(url_root)modules/safeguards.cmake'
       - '%(url_root)modules/version.cmake'
--- a/config.h.in
+++ b/config.h.in
@@ -6,6 +6,8 @@
 
 #define BLAS_H "@BLAS_H@"
 
+#cmakedefine HAVE_MPI
+
 #cmakedefine HAVE_OPENMP
 
 #cmakedefine HAVE_BLAS
--- a/setup
+++ b/setup
@@ -20,6 +20,7 @@ Usage:
 Options:
   --cxx=<CXX>                            C++ compiler [default: g++].
   --extra-cxx-flags=<EXTRA_CXXFLAGS>     Extra C++ compiler flags [default: ''].
+  --mpi                                  Enable MPI parallelization [default: False].
   --omp                                  Enable OpenMP parallelization [default: False].
   --coverage                             Enable code coverage [default: False].
   --enable-tests                         Enable tests [default: False].
@@ -42,6 +43,7 @@ def gen_cmake_command(options, arguments):
     command.append('CXX={0}'.format(arguments['--cxx']))
     command.append(arguments['--cmake-executable'])
     command.append('-DEXTRA_CXXFLAGS="{0}"'.format(arguments['--extra-cxx-flags']))
+    command.append('-DENABLE_MPI={0}'.format(arguments['--mpi']))
     command.append('-DENABLE_OPENMP={0}'.format(arguments['--omp']))
     command.append('-DENABLE_CODE_COVERAGE={0}'.format(arguments['--coverage']))
     command.append('-DENABLE_TESTS={0}'.format(arguments['--enable-tests']))
--- a/src/FunctionTree.cpp
+++ b/src/FunctionTree.cpp
@@ -22,18 +22,6 @@ FunctionTree<D>::FunctionTree(const MultiResolutionAnalysis<D> &mra)
     this->resetEndNodeTable();
 }
 
-/** FunctionTree constructor for Serial Tree using shared memory.
-  * */
-template<int D>
-FunctionTree<D>::FunctionTree(const MultiResolutionAnalysis<D> &mra, SharedMemory* &shMem)
-        : MWTree<D> (mra) {
-    this->serialTree_p = new SerialFunctionTree<D>(this);
-    this->serialTree_p->isShared = true;
-    this->serialTree_p->shMem = shMem;    
-    this->serialTree_p->allocRoots(*this);
-    this->resetEndNodeTable();
-}
-
 //** FunctionTree destructor. */
 template<int D>
 FunctionTree<D>::~FunctionTree() {
--- a/src/FunctionTree.cpp
+++ b/src/FunctionTree.cpp
@@ -15,9 +15,9 @@ using namespace Eigen;
 /** FunctionTree constructor for Serial Tree.
   * */
 template<int D>
-FunctionTree<D>::FunctionTree(const MultiResolutionAnalysis<D> &mra)
+FunctionTree<D>::FunctionTree(const MultiResolutionAnalysis<D> &mra, SharedMemory *sh_mem)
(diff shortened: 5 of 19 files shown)


gitpeterwind:
I do not understand where you want to go. 
This change will remove the possibility to use (and develop) MPI parallelism, or do I understand it wrong? 
This is then not an "Add simple MPI features", but rather "remove MPI parallelism".

bast:
The way I understand it is that MPI layer is being moved higher "up" - is that right?

gitpeterwind:
Can the code still do a MPI parallel run after this change?

stigrj:
This is an attempt to make a clean cut between mrchem and mrcpp, where the MPI part is the only thing that is currently tangled up between the two. With this patch I define what will be on the mrcpp side, everything else will be on the mrchem side. The mrcpp API should later be updated with non-blocking communication and shared memory features, but I think we should remove all internal pointers in the data structures first to make this part more transparent.

Everything that is now "removed" from the MPI part will be added to the mrchem side when the "new" mrcpp is incorporated into mrchem (this is not yet done). This will require some minor tweaking in order to fit with the new API, but every implemented MPI feature should survive, except the shared memory part will will be added later when this feature is available through mrcpp.

gitpeterwind:
ok. I find it dangerous to remove features and "plan to put them back in a better form". It is safer to follow a path where nothing is removed before something better is in place.
We will soon need other MPI features, such as communicator groups (pointing to "molecules" or "fragments"); this will be more difficult if mrcpp does not allow for this kind of flexibility

bast:
Just offering a counter-argument: preserving features makes it harder to add new/better ones. But I also see your point Peter.

gitpeterwind:
as long as we (temporarily) keep the mrcpp under mrchem with MPI until a working interface is in place, it is not problematic to have the standalone version in addition.
It would be nice to remove the "internal pointers" from the nodes, yes, but need some deep changes. We can discuss this.

stigrj:
I share your concern about removing features. There are things I removed from mrchem two years ago that are still not re-introduced because I "just have to fix this and that first" (important stuff, like the entire response code and open-shell DFT)... I guess the solution is that we need to upgrade mrcpp with the features we need sooner rather than later, and until then keep the old mrcpp in mrchem.

Regarding the communicators, I don't think the current API lacks the flexibility for adding other groups on top. At least it is my intention that this should be possible, but I might not see all implications of my actions here :)

gitpeterwind:
I can try to start to remove pointers from nodes. Some of the work is already done: in "rewrite pointers" the parent and children pointers are actually recalculated using the index. One could just move that calculation into "getParent" and "getChildren".  To remove the pointer to own tree is more difficult (but possible).
If I integrate my changes in the mrcpp from mrchem (where I know better how to test), would it be easy to port to the standalone mrcpp?

gitpeterwind:
> I can try to start to remove pointers from nodes.

I have a version that removed the pointers to child, parent and coefficients. It pass all test for omp=1, also mpi run for h2o runs fine. However the test [identity_convolution] fails sometimes (but not always) for omp>1. 
The modifications, in principle, should not affect the omp parts, but we should discuss this. (it could be related to issue #5  )
The modifications were in principle easy (add a few lines in getChild, GetParent and GetCoefs), but turned out to be much more complex. These few lines had to be placed in many classes and under several "if". MWNodes, GenNodes, ProjectedNodes, OperatorNodes each needed som kind of special treatment.
Other problems:  the code got about 10% slower (omp=1), which is not negligible. 
There is an additional problem ahead when removing the "omp_lock_t" as this is from the omp library and contains probably pointers.
Now I am getting pessimistic about doing something nice here. We should discuss what is really needed as this has a price in terms of efficiency and code complexity.... and programming time.

stigrj:
Made some changes to the MPI interface, bringing it closer to the original mrchem syntax. Non-blocking ``isend`` and shared memory are re-introduced. Hopefully this will provide enough flexibility to be re-incorporated into mrchem without losing any features.

New API
------------

An MPI communicator must be initialized (and finalized) by the host program:

```cpp
int rank, size;
MPI_Init(&argc, &argv);
MPI_Comm_rank(MPI_COMM_WORLD, &rank);  // Get my rank ID
MPI_Comm_size(MPI_COMM_WORLD, &size);  // Get total number of ranks
MPI_Finalize();
```
Work distribution is achieved brute force by assigning different functions to particular ranks:
```cpp
FunctionTree<3> tree(MRA);       // All ranks define the tree
if (rank == 0) project(tree, f); // Only rank 0 projects the function
```
A ``FunctionTree`` can be sent between MPI ranks using blocking communication :
```cpp
int src=0, dst=1; // Source and destination ranks
int tag = 111111; // Unique tag for each communication
if (rank == src) send_tree(tree, dst, tag, MPI_COMM_WORLD);
if (rank == dst) recv_tree(tree, src, tag, MPI_COMM_WORLD);
```
Or using non-blocking send (non-blocking receive is not supported):
```cpp
int src=0, dst=1; // Source and destination ranks
int tag = 222222; // Unique tag for each communication
MPI_Request request = MPI_REQUEST_NULL;
if (rank == src) isend_tree(tree, dst, tag, MPI_COMM_WORLD, &request);
if (rank == dst) revc_tree(tree, src, tag, MPI_COMM_WORLD);

// Source rank waits until the send is complete
MPI_Wait(&request, MPI_STATUS_NULL);
```

Finally, a ``FunctionTree`` can be shared among MPI processes that share the same physical memory space. In this case a new shared memory communicator is required in the host program:
```cpp
MPI_Comm MPI_COMM_SHARE;
MPI_Comm_split_type(MPI_COMM_WORLD, MPI_COMM_TYPE_SHARED, 0, MPI_INFO_NULL, &MPI_COMM_SHARE);
```
The ``FunctionTree`` can be assigned to the shared memory space using a ``SharedMemory`` object:
```cpp
int mem_size = 1000; // MB
SharedMemory shared_mem(MPI_COMM_SHARE, mem_size);
FunctionTree<3> tree(MRA, shared_mem);
```
Now the tree can be projected by one of the ranks with write access:
```cpp
if (rank == 0) project(tree, f); // Only rank 0 (of shared group) projects
```
Finally, the shared tree must be "synchronized" between the sharing ranks. The reason for this is that the internal memory pointers needs to be updated *locally* on each MPI process whenever the shared memory window has been updated.
```cpp
int src = 0;      // Source rank in shared group
int tag = 333333; // Unique tag for each communication
share_tree(tree, src, tag, MPI_COMM_SHARE);
```
This means that ``share_tree`` must be called every time a shared ``FunctionTree`` is changed.

bast:
A comment/question: should the code really work with MPI_COMM_WORLD? How about only assuming the code gets *a* communicator which might be the MPI_COMM_WORLD or a subset, but you let the caller decide?

stigrj:
@bast, I think I was a bit unclear in my comment, as Peter had the same concern :smiley:  The part with the communicators are added only as examples of how to use it. There is no reference to either MPI_COMM_WORLD or MPI_COMM_SHARE inside MRCPP, the communicator is passed as argument to all functions, and any communicator will be accepted. I will make this more clear in the _official_ documentation.

bast:
OK very good. I think your doc is good and it is good that we agree on that point.

PR #11, opened 2017-12-20
New API

stigrj:
This is a massive pull request that defines and documents the API of MRCPP. A lot of files and lines have changed, but looking at individual commits you will find that most of the changes are very similar. I will describe the main ideas below, but please download and build the documentation locally for more info. Input and suggestions are very welcome!

MRCPP include files
---------------------------
The necessary header files for the API have been grouped into seven different files that can be included by the application program:
* MRCPP/MWFunctions
* MRCPP/MWOperators
* MRCPP/Gaussians
* MRCPP/Parallel
* MRCPP/Printer
* MRCPP/Plotter
* MRCPP/Timer

Everything within MRCPP is now enclosed in a new ``mrcpp`` namespace. I removed Timer.h and Printer.h from all header files, to avoid them being unintentionally included. This required a lot of new source files since many classes that used to be header-only included one or both of these headers.

Tree builders
----------------
The old TreeBuilder classes have been replaced by standalone functions. The reason for this change is simplicity, there's no need for these things to be classes, everything is contained in simple functions. Now we don't have to construct the builder before we can use it.
* MWProjector -> mrcpp::project()
* MWAdder ->mrcpp::add()
* MWMultiplier -> mrcpp::multiply()
* MWConvolution -> mrcpp::apply()
* MWDerivative -> mrcpp::apply()
* GridGenerator -> mrcpp::build_grid() and mrcpp::copy_grid()
* GridCleaner -> mrcpp::clear_grid()

Please have a look in the ``examples`` directory to see how it works.

Pilot code
-------------
The examples in the old pilot code have been moved to an``examples`` directory, where each example is compiled separately. Also added some MPI examples here. The ``pilot`` directory remains as a sandbox for experimentation, but the sample code snippet is now a clean sheet.


ilfreddy:
How urgent is it to approve this one? Given the extent of the change, I think it would be very pedagogic to go through it in the first MRChem meeting after Xmas.

stigrj:
No rush, I switched to vacation mode this afternoon :santa: :christmas_tree:

juselius:
Bravo!

PR #15, opened 2018-01-24
Implement saveTree and loadTree for FunctionTrees

stigrj:
FunctionTrees can now be saved as binary files
and loaded at a later stage or in subsequent runs.
Implicitly assumes that the MRA of the tree which
is loaded into is identical to the saved tree.
The serialization is similar to MPI send/recv.

### DIFF
--- a/src/FunctionTree.cpp
+++ b/src/FunctionTree.cpp
@@ -72,7 +72,7 @@ void FunctionTree<D>::saveTree(const string &file) {
     SerialFunctionTree<D> &sTree = *this->getSerialFunctionTree();
 
     // Write size of tree
-    int nChunks = sTree.nodeChunks.size();
+    int nChunks = sTree.getNChunksUsed();
     f.write((char*) &nChunks, sizeof(int));
 
     // Write tree data, chunk by chunk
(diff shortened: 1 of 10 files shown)


stigrj:
How do you know which chunks are valid?

gitpeterwind:
There are several possibilities:
1) Look at sTree.lastNode. The chunk it is on, is the last used.
2) look at nodeStackStatus. Since you already loop over them it is perhaps the easiest: if no  nodeStackStatus is 1 one a chunk, it is empty and does not need to be stored.
3) Implement a new flag that shows i a chunk is empty or not (which is set when written on, unset when all nodes on it are deleted).

stigrj:
Do you already have a way of clearing a tree while keeping the allocated memory? With clearing I mean the tree returning to its initial state containing only root nodes with uninitialized coefs. If not then I will write one.

stigrj:
It is supposed to be the ``tree.clear()`` function, which is currently NOT_IMPLEMENTED_ABORT, but you might have implemented a similar feature somewhere else?

stigrj:
Turned out to be very simple, just delete all children of root nodes, and clear the coefs/norms of the roots. Now only used chunks are saved to disk (and sent through MPI).

PR #17, opened 2018-03-24
Fixing some issues encountered when compiling the new MRChem

ilfreddy:
the doc folder does not contain a CMakeLists.txt because the documentation (in MRChem) is now built with a script. Assuming we want the same for mrcpp, I have now removed the doxygen line from the main CMakeLists.txt


ilfreddy:
Removed doxygen from autocmake as well.
I also get another error when compiling the new mrchem. The compiler requires all template instantiations to carry the `mrcpp::` namespace.

stigrj:
All of this should be incorporated in `master` after PR #26

PR #18, opened 2018-04-07
Cleanup source

stigrj:
Moved around some source and header files. Final layout: all sources and *regular* header files are located under `src`and logically grouped into sub folders:
```
src/mwfunctions
src/mwoperators
src/mwbuilders
src/mwutils
src/mwcore
src/mwtrees
```

 The API headers are now located in the `api`folder (currently these will simply `#include` batches of regular headers):
```
api/MWFunctions
api/MWOperators
api/Gaussians
api/Printer
api/Plotter
api/Timer
```

With `make` all headers are copied to `build/include/MRCPP`, with the *regular* headers one level further under `src`:
```
build/include/MRCPP/MWFunctions
build/include/MRCPP/MWOperators
build/include/MRCPP/Gaussians
build/include/MRCPP/Printer
build/include/MRCPP/Plotter
build/include/MRCPP/Timer
build/include/MRCPP/src/*.h
```

With `make install` the headers will be installed in the same way under `install_path`. The library will *not* get the `MRCPP` prefix:
```
install_dir/include/MRCPP/...
install_dir/lib/libmrcpp.a
```


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) 2015-2018 by Radovan Bast, Roberto Di Remigio, Jonas Juselius, and contributors.
 
 # set minimum cmake version
-cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.3 FATAL_ERROR)
 
 # project name
 project(MRCPP CXX)
```
robertodr: Consider using the extended syntax:
```
project(MRCPP LANGUAGES CXX VERSION ...)
```
This initialized all the `PROJECT` related options consistently. The latest Autocmake implementes the `LANGUAGES` option, so you should update to that one.

api/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+    MWFunctions
+    MWOperators
+    Gaussians
+    Parallel
+    Plotter
+    Printer
+    Timer
+    )
+
+# Sets install directory for all the headers in the list
```
robertodr: The copy and install can be done more compactly without the loop:
```
file(
  COPY ${CMAKE_CURRENT_LIST_DIR}
   DESTINATION ${PROJECT_BINARY_DIR]/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}
 )
install(
   DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}
  )
```
but please test that is works as expected!

cmake/custom/catch.cmake
```diff
@@ -1 +1 @@
-include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/external/Catch)
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/external/catch)
```
robertodr: If you are using Catch2, you can actually remove this file and to what is suggested on their docs. Put this in the `CMakeLists.txt` under the `tests` directory:
```
# Prepare "Catch" library for other executables
set(CATCH_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/catch)
add_library(Catch INTERFACE)
target_include_directories(Catch INTERFACE ${CATCH_INCLUDE_DIR})
```
and then link the unit tests executable as:
```
target_link_libraries(tests Catch)
```
The benefit is that everything test-related is localized to where the tests actually are and the rest of the project doesn't have the spurious include in their compiler flags.

cmake/custom/eigen.cmake
```diff
@@ -1 +1,2 @@
-find_package(Eigen3 REQUIRED)
+find_package(Eigen3 3.1.0 REQUIRED)
```
robertodr: You could use the [imported target](http://eigen.tuxfamily.org/dox/TopicCMakeGuide.html) for Eigen (you'll need to upgrade to 3.3.0 though). And then instead of having the global `include_directories` here you can have `target_link_libraries(example Eigen3::Eigen)` wherever needed.

cmake/custom/examples.cmake
```diff
@@ -1,3 +1,10 @@
+#.rst:
+#
+# autocmake.yml configuration::
+#
+#   docopt: "--enable-examples Enable code examples [default: False]."
+#   define: "'-DENABLE_EXAMPLES={0}'.format(arguments['--enable-examples'])"
+
```
robertodr: This variable might be unbound/undefined in case a user decides to ditch the `setup` script.
You should try is to configure the project _without_ using the `setup` script, for example:
`cmake -H. -Bbuild` will configure the project to be built into the `build` directory. Then all unbound/undefined variables will cause trouble or lead to nonsensical configuration. In the latest Autocmake we added a couple of macros to remedy this issue. Here you should add a:
```
option_with_print(ENABLE_EXAMPLES "Enable code examples" OFF)
```
which will set it to off if `ENABLE_EXAMPLES` was not specified, ensuring that the variable is always in a well-defined state.

cmake/custom/main.cmake
```diff
@@ -5,12 +5,13 @@ configure_file (
     "${PROJECT_BINARY_DIR}/config.h"
     )
 
-include_directories(${PROJECT_BINARY_DIR})
-
-include_directories(${PROJECT_BINARY_DIR}/external/include)
-link_directories(${PROJECT_BINARY_DIR}/external/lib)
+install(
+    FILES ${CMAKE_BINARY_DIR}/config.h
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/src
```
robertodr: I think the destination of the `config.h` header should be just `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}`

cmake/custom/tests.cmake
```diff
@@ -1,6 +1,13 @@
+#.rst:
+#
+# autocmake.yml configuration::
+#
+#   docopt: "--enable-tests Enable tests [default: False]."
+#   define: "'-DENABLE_TESTS={0}'.format(arguments['--enable-tests'])"
+
 option(ENABLE_TESTS "Enable test suite" ON)
```
robertodr: Consider using the `option_with_print` function from Autocmake (see previous comment)

cmake/custom/tests.cmake
```diff
@@ -1,6 +1,13 @@
...
 
-macro(add_Catch_test _name _labels)
+macro(add_catch_test _name _labels)
```
robertodr: I've rewritten this macro for [PCMSolver](https://github.com/PCMSolver/pcmsolver/blob/release/1.2.Z/cmake/custom/test.cmake) It now only works if you give named arguments, which helps readability:
```
add_Catch_test(
  NAME
    bi_operators_purisima
  LABELS
    bi_operators
    bi_operators_purisima
  REFERENCE_FILES
    uniformdielectric_D_purisima.npy
    vacuum_D_purisima.npy
  )
```
Feel free to copy and adapt at will.

src/CMakeLists.txt
```diff
@@ -1,72 +1,45 @@
...
+    constants.h
+    eigen_disable_warnings.h
+    macros.h
+    mrcpp_declarations.h
+    parallel_mpi.h
+    parallel_omp.h
+    )                                                       
+                                                                                 
+# Sets install directory for all the headers in the list                         
+foreach(_header ${mrcpp_h})
```
robertodr: I think these files should go under `${CMAKE_INSTALL_INCLUDEDIR]/${PROJECT_NAME}` You might remove the loop using the `REGEX` option to `file(COPY` and `install` but you'll have to experiment a bit since I've never tried it.

src/mwbuilders/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/WaveletAdaptor.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/add.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/apply.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/grid.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/multiply.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/project.h
+    )                                                       
+                                                                                 
+# Sets install directory for all the headers in the list                         
+foreach(_header ${mwbuilders_h})
```
robertodr: I think these headers should go under `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/mwbuilders`

src/mwcore/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/MWFilter.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/ObjectCache.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/QuadratureCache.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/ScalingBasis.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/ScalingCache.h
+    )                                                       
+                                                                                 
+# Sets install directory for all the headers in the list                         
+foreach(_header ${mwcore_h})                                                 
+ file(
```
robertodr: Same here should be `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/mwcore`

src/mwfunctions/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/GaussFunc.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/GaussPoly.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/Gaussian.h
+    )                                                       
+                                                                                 
+# Sets install directory for all the headers in the list                         
+foreach(_header ${mwfunctions_h})                                                 
+ file(
+    COPY ${_header}
+    DESTINATION ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/src
```
robertodr: Same here should be `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/mwfunctions`

src/mwoperators/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/PHOperator.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/PoissonKernel.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/PoissonOperator.h
+    )                                                       
+                                                                                 
+# Sets install directory for all the headers in the list                         
+foreach(_header ${mwoperators_h})                                                 
+ file(
+    COPY ${_header}
+    DESTINATION ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/src
```
robertodr: Same here should be `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/mwoperators`

src/mwoperators/DerivativeKernel.h
```diff
@@ -21,6 +21,8 @@ class DerivativeKernel : public GreensKernel {
         GaussPoly<1> dg = g.differentiate(0);
         this->append(dg);
     }
+
+    virtual std::ostream& print(std::ostream &o) const { return o; }
```
robertodr: Isn't it better to just overload `operator<<` instead of passing a stream and returning it?

src/mwtrees/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/SerialOperatorTree.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/SerialTree.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/TreeIterator.h
+    )                                                       
+                                                                                 
+# Sets install directory for all the headers in the list                         
+foreach(_header ${mwtrees_h})                                                 
+ file(
+    COPY ${_header}
+    DESTINATION ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/src
```
robertodr: Same here should be `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/mwtrees`

src/mwtrees/MWNode.cpp
```diff
@@ -23,11 +23,11 @@ template<int D>
 MWNode<D>::MWNode()
         : tree(0),
           parent(0),
-          nodeIndex(),
-          hilbertPath(),
           squareNorm(-1.0),
-          n_coefs(0),
           coefs(0),
```
robertodr: You should use [`nullptr`](http://en.cppreference.com/w/cpp/language/nullptr) rather than `0` to initialize the `double * coefs` data member because `nullptr` is standard for the null pointer whereas `0` can be interpreted as an integer.

robertodr: Same probably goes for the other pointer members of this class. Is the use of `std::unique_ptr` an option here? The syntax and semantics are the same as those of a raw pointer and since `unique_ptr` doesn't have the reference counting of `shared_ptr` you don't get the overhead.

src/mwutils/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/Plotter.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/Printer.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/Timer.h
+    )                                                       
+                                                                                 
+# Sets install directory for all the headers in the list                         
+foreach(_header ${mwutils_h})                                                 
+ file(
+    COPY ${_header}
+    DESTINATION ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/src
```
robertodr: I think this should be `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/mwutils`

cmake/custom/examples.cmake
stigrj: Aha, is this also the reason this option does not show up in the cmake gui?

robertodr: Yep. It is never declared as an `option`.

src/mwoperators/DerivativeKernel.h
stigrj: I took this from your coding standards :stuck_out_tongue: `operator<<` is overloaded in the base class which calls a virtual `print()` function that is implemented in the derived classes. Here I just added a quiet print since it was a missing pure virtual and Luca's compiler complained (not mine for some reason...).

robertodr: LOL. Can you add a comment that this is a silent print? The implementation is a bit confusing :) BTW what was the compiler warning/error?

stigrj:
Thanks for the thorough review. Does the layout of the installed `include` files always have to be identical to the actual `src` tree? My idea was to make it as simple as possible, but it doesn't matter much if we get an extra layer of folders. Also, should *all* headers of the project be installed, or just the ones meant to be used in the API?

api/CMakeLists.txt
stigrj: These commands work, but the api files now goes to `include/MRCPP/api/`. Is this how it has to be, or  can I circumvent the `api` layer in some way?

stigrj:
Should it not be a `src` in the install path? `${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/src/mwtrees/something.h` etc.? Actually, it would be better without, if the application program could include the api headers as `#include "MRCPP/MWFunctions"`, and in case you need something special `#include "MRCPP/mwtrees/SomethingSpecial.h"`, but then I will move all sources into a subfolder of `src`, so that the only visible files under `MRCPP`are the api headers and `config.h`.

api/CMakeLists.txt
robertodr: If you try with the `${CMAKE_CURRENT_LIST_DIR}/*`? If that doesn't work disregard this comment altogether.

robertodr:
No there shouldn't be `src` in the install path. The situation you want to have is the one you just described. You want to be able to program with those headers from within their repository and from an installed location in the exact same way.

stigrj:
Awsomeness! I will give it a try

stigrj:
Moved filters to `share` and some headers to `include`. Headers are now installed to `install-dir/MRCPP/include/mwtrees` etc. and the includes within the project are now explicit `#include "mwtrees/something.h"`.

cmake/custom/catch.cmake
stigrj: I will upgrade Catch and do these changes in a separate PR

cmake/custom/eigen.cmake
stigrj: I will incorporate this with https://github.com/MRChemSoft/mrchem/issues/79, currently we cannot use Eigen3.3 on stallo and travis

cmake/custom/tests.cmake
stigrj: This will also be updated in #22

src/mwoperators/DerivativeKernel.h
stigrj: Implemented a proper print instead

src/mwtrees/MWNode.cpp
stigrj: I will address this separately, #19

stigrj:
Ready to go, please approve or request further changes

PR #27, opened 2018-04-16
MathUtils -> math_utils

stigrj:
* Change `MathUtils` class (that used to contain only `static` functions) into `math_utils` namespace.
* Remove unused `math_utils` functions. Some of them will pop up on the mrchem side.
* More consistent naming: `math_utils`, `omp_utils`, `mpi_utils`

- [x] Ready to go


src/mwutils/mwmath.cpp
```diff
@@ -18,12 +17,10 @@ using namespace Eigen;
 using namespace mrcpp;
 
 /** Calculate \f$ m^e\f$ for integers. */
-int MathUtils::ipow(int m, int e) {
-    if (e < 0) {
-        MSG_FATAL("Exponent cannot be negative: " << e)
-    }
+int mwmath::ipow(int m, int e) {
```
robertodr: You could probaly use `size_t` or `unsigned int` to have the compiler check for you that an positive integer was passed. Also a `static_assert` rather than a runtime check might be better here. Out of curiosity, does this outperform `std::pow`?

src/mwutils/mwmath.cpp
```diff
@@ -37,14 +34,14 @@ int MathUtils::ipow(int m, int e) {
  *	The norm of the matrix is obtained as:
  *	 \f$ ||M|| \lim_{n \rightarrow \infty} ||x_n||/||x_{n-1}||\f$
  */
-double MathUtils::matrixNorm2(const VectorXd &vector) {
-    int n = (int) sqrt(vector.size());
+double mwmath::matrixNorm2(const VectorXd &vector) {
```
robertodr: You might want to explore this https://eigen.tuxfamily.org/dox/group__TutorialReductionsVisitorsBroadcasting.html
Also, be aware that passing Eigen types as parameters might impact performance negatively, if not done correctly: https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html

src/mwutils/mwmath.cpp
stigrj: I guess the purpose of this `ipow` is that it returns an `int`, all versions of `std::pow` returns `floats`, right?

robertodr: Yes, http://en.cppreference.com/w/cpp/numeric/math/pow
Out of curiosity, where is this used?

stigrj: It's mostly used to compute `(k+1)^D`, where `1 < k < 20` is the polynomial order, and `1 < D < 3` is the dimension. Could easily be replaced by inline evaluations like
```
int result = (k+1)*(k+1);
```
or
```
int result = 1;
for (int d = 0; d < D; d++) result *= (k+1);
```

stigrj: But in the end this `ipow`is not performance critical, it's just for convenience.

robertodr: Yeah, sorry. It just piqued my interest :)

src/mwutils/mwmath.cpp
stigrj: I changed the matrix norms to the `lpNorm` routines of Eigen, but it turns out they require v.3.3. Is it possible to get this version on Travis without shipping our own  Eigen? Currently I get v3.2 when installing `libeigen3-dev` on Travis, while I get v3.3.4 on my laptop.

stigrj: For the Eigen types, I guess we are fine as long av we pass by `const` reference. This is done in the most performance critical routines at least.

robertodr: Add something like this to `.travis.yml`:
```
  cd $HOME/Downloads
  mkdir -p eigen
  curl -Ls http://bitbucket.org/eigen/eigen/get/${Eigen_VERSION}.tar.gz | tar -xz -C eigen --strip-components=1
  cd eigen
  cmake -H. -Bbuild_eigen -DCMAKE_INSTALL_PREFIX=$HOME/Deps/eigen &> /dev/null
  cmake --build build_eigen -- install &> /dev/null
  cd $TRAVIS_BUILD_DIR
  rm -rf $HOME/Downloads/eigen
```
and then when configuring the project add:
```
--cmake-options="-DEigen3_DIR: '$HOME/Deps/eigen/share/eigen3/cmake/"
```
The `find_package` for Eigen should be:
```
find_package(Eigen3 3.3 REQUIRED NO_MODULE)
```
and then you can do:
```
target_link_libraries(...
  PUBLIC
    Eigen3::Eigen
  )
```

.travis.yml
```diff
@@ -51,14 +51,21 @@ matrix:
         - PYTHON_VER='2.7'
 
 before_script:
+  - cd $HOME/Downloads
+  - mkdir -p eigen
+  - curl -Ls http://bitbucket.org/eigen/eigen/get/${Eigen_VERSION}.tar.gz | tar -xz -C eigen --strip-components=1
```
robertodr: I think you forgot to define `Eigen_VERSION` :grin:

stigrj: Thought it was pick up from `cmake` :stuck_out_tongue:  It failed on `cd $HOME/Downloads` though...

stigrj:
@robertodr can you foresee any conflict here with your ongoing work? In particular regarding Eigen

robertodr:
No, I'll be working exclusively on MRChem. Don't let my slow progress stop you. In case there are conflicts, I'll fix them or ask for help :wink:

stigrj:
Then I would appreciate a positive review :wink:

.travis.yml
```diff
@@ -14,45 +14,52 @@ matrix:
...
+  - cd $HOME/Downloads
+  - curl -Ls http://bitbucket.org/eigen/eigen/get/${EIGEN_VER}.tar.gz | tar -xz -C eigen --strip-components=1
+  - cd eigen
+  - cmake -H. -Bbuild_eigen -DCMAKE_INSTALL_PREFIX=$HOME/Deps/eigen &> /dev/null
+  - cmake --build build_eigen -- install &> /dev/null
   - cd ${TRAVIS_BUILD_DIR}
   - python --version
   - cmake --version
   - ${CXX_COMPILER} --version
+  - rm -rf $HOME/Downloads/eigen
```
robertodr: Instead of removing it every time you can use the [cache on Travis](https://docs.travis-ci.com/user/caching/#Arbitrary-directories). You won't spend time downloading and installing. Look also [here](https://github.com/bast/cmake-recipes/blob/master/testing/dependencies/install_dependencies.sh#L58-L73) and [here](https://github.com/bast/cmake-recipes/blob/master/.travis.yml#L59-L67)

cmake/custom/eigen.cmake
```diff
@@ -1,2 +1,2 @@
-find_package(Eigen3 3.1.0 REQUIRED)
+find_package(Eigen3 3.3 REQUIRED NO_MODULE)
```
robertodr: You can use `CONFIG` which is synonymous with `NO_MODULE` and is more canonical.

cmake/custom/eigen.cmake
```diff
@@ -1,2 +1,2 @@
...
 include_directories(SYSTEM ${EIGEN3_INCLUDE_DIR})
```
robertodr: This is not needed anymore.

cmake/custom/eigen.cmake
robertodr: Also, could you add a little message for the user:
```
if(TARGET Eigen3::Eigen)
  message(STATUS "Eigen3 version ${EIGEN3_VERSION_STRING} found in ${EIGEN3_INCLUDE_DIR}")
endif()
```

cmake/custom/eigen.cmake
stigrj: Some strange behavior:


It works with
```
include_directories(${EIGEN3_INCLUDE_DIR})
```
but not with 
```
target_include_directories(mrcpp-shared PUBLIC ${EIGEN3_INCLUDE_DIR})
target_include_directories(mrcpp-static PUBLIC ${EIGEN3_INCLUDE_DIR})
```
and not with
```
target_link_libraries(mrcpp-shared PUBLIC Eigen3::Eigen)
target_link_libraries(mrcpp-static PUBLIC Eigen3::Eigen)
```

robertodr: OK, so the dependency has moved down to `mrcpp-objlib`:
```
target_include_directories(mrcpp-objlib
  PRIVATE
    ${PROJECT_SOURCE_DIR}/include
    ${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}
    ${CMAKE_CURRENT_LIST_DIR}
  PUBLIC
    ${EIGEN3_INCLUDE_DIR}
  )
```
The `OBJECT` library builds the source files into object files, without linking them. So you get the object files and can decide if you want to link them statically or dynamically or both (as we do now). Dependencies move down from the library targets to the `OBJECT` library. Hence your error above.

A couple of letdowns:
- `target_link_libraries` can't be used with `OBJECT` libraries (it probably can with a more recent version of CMake, but it doesn't matter for us)
- The target exported by Eigen cannot be used with `target_include_directories` hence the use of the variable `EIGEN3_INCLUDE_DIR` defined by the same package...

robertodr: Alas, to get tests compiling you'll also need to add `${EIGEN3_INCLUDE_DIR}` to the `INTERFACE` include directories for `mrcpp-shared` and `mrcpp-static`[^1] (again, pity that Eigen as imported target is flaky...)
In `tests/CMakeLists.txt` you can remove
```
target_include_directories(mrcpp-tests
  PRIVATE
    ${PROJECT_BINARY_DIR}/include/MRCPP
  )
```

[^1] I am looking for a way to make `src/CMakeLists.txt` more compact, because now it's a bit bloated and repetitive, I must admit.

stigrj:
Can someone please look at this? @bjorgve? @ilfreddy?

PR #28, opened 2018-04-19
[WIP] Enable build static and shared libraries both

robertodr:
- Tests are enabled by default.
- Executables are put in `build/bin`, libraries in `build/lib` or `build/lib64`.
- Sources are compiled as object files and then linked into static and shared libraries.


stigrj:
How do you choose which library to link to? By the size of the example executables they seem to be dynamically linked, but I see nothing in `examples/CMakeLists.txt` that would indicate one or the other.

cmake/custom/static_library.cmake
```diff
@@ -0,0 +1,26 @@
...
+#
+# Enables creation of static/shared library.
+#
+# autocmake.yml configuration::
+#
+#   docopt: "--static Create only the static library [default: False]."
+#   define: "'-DSTATIC_LIBRARY_ONLY={0}'.format(arguments['--static'])"
+
+
+option_with_print(
```
stigrj: Does the `option_with_print` work as expected now?

robertodr: Settings go through correctly, but the value printed to screen is always the one set as default. I am working on a fix Autocmake-side.

PR #30, opened 2018-05-03
Add some minor functionality to FunctionTreeVector

ilfreddy:
I upgraded `FunctionTreeVector`with some minor functionality:
1) `push_back` a whole `FunctionTreeVector`
2) count the total number of nodes.

The functionality is IMO generic enough, and in particular it allows me to simplify `XCFunctional` and to remove some duplicate stuff there.

### DIFF
--- a/src/trees/FunctionTreeVector.h
+++ b/src/trees/FunctionTreeVector.h
@@ -17,7 +17,7 @@ public:
 
     int size() const { return this->funcs.size(); }
     void clear(bool dealloc = false);
-    int sumNodes();
+    int sumNodes() const;
 
     void push_back(double c, FunctionTree<D> *f);
     void push_back(FunctionTree<D> *f);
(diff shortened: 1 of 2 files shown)


robertodr:
Would it make sense to implement this using STL containers? This seems to be a `std::vector<std::tuple<double, FunctionTree<D> *>>` You'd probably get a lot of the functions implemented for free.

stigrj:
Good idea @robertodr

src/trees/FunctionTreeVector.h
```diff
@@ -17,9 +17,11 @@ class FunctionTreeVector {
 
     int size() const { return this->funcs.size(); }
     void clear(bool dealloc = false);
+    int sumNodes();
```
stigrj: I think you can add a trailing `const` on this one

stigrj:
I'm squashing this one

PR #31, opened 2018-05-06
Some fixes for the include directories

robertodr:
This PR reshuffles a bit the inclusion of headers so that when MRCPP is consumed as imported target everything goes without a hitch. I've refined this working on making the build of MRChem a superbuild (PR coming up for that)
- Inclusion of headers _within_ the same subdirectory are not prepended by the name of the subdirectory, _e.g._ `core/CrossCorrelationCache.h` does `#include "CrossCorrelation.h"` and not `#include "core/CrossCorrelation.h"`.
- `${CMAKE_INSTALL_INCLUDEDIR}/MRCPP` it's now in the `INSTALL_INTERFACE` This means that a consumer will get `-I${CMAKE_INSTALL_INCLUDEDIR} -I${CMAKE_INSTALL_INCLUDEDIR}/MRCPP` in its include line. I think only the first part is necessary, but then `config.h` & co aren't picked up correctly. These should probably be included internally as `MRCPP/config.h` and so on, but that would have meant making quite some radical changes to the current state of affairs and I leave the decision to  you whether to make those or not.


robertodr:
Well, of course I focked up the examples...

PR #33, opened 2018-05-15
Rewrite FunctionTreeVector<D>

robertodr:
Rewrite `FunctionVectorTree<D>` as `std::vector<std::tuple<double, FunctionTree<D> *>>`. This change snowballed a bit, but I think I addressed all ramifications. Notable changes (which you might not agree with, so feel free to disregard this PR entirely):
  - `push_back` requires to **explicitly** create a coefficient-function tuple:
     ```
    tree_vec.push_back(std::make_tuple(coef, func));
     ```
     rather than:
     ``` 
     tree_vec.push_back(coef, func);
     ```
     It's probably possible to specialize `push_back` to keep the old behavior.
  - `getCoef` and `getFunc` are now free functions accepting `FunctionVectorTree<D>` as first argument.


src/treebuilders/multiply.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
-#include "mrcpp_declarations.h"
+#include "functions/RepresentableFunction.h"
+#include "trees/FunctionTree.h"
+#include "trees/FunctionTreeVector.h"
```
stigrj: Any reason to include rather than forward declare?

src/operators/ConvolutionOperator.cpp
```diff
@@ -55,17 +56,15 @@ void ConvolutionOperator<D>::initializeOperator(GreensKernel &greens_kernel) {
         println(10, "Time transform      " << trans_t);
         println(10, std::endl);
 
-        this->kern_exp.push_back(k_tree);
+        this->kern_exp.push_back(std::make_tuple(1.0, k_tree));
```
stigrj: I'm not crazy about this syntax. Is it important to make it clear that the (coef, tree) forms a tuple under the hood?

src/trees/FunctionTreeVector.h
```diff
@@ -1,38 +1,53 @@
...
+     }
+   }
+   fs.clear();
+}
 
-    double getCoef(int i) const;
-    FunctionTree<D> &getFunc(int i);
-    const FunctionTree<D> &getFunc(int i) const;
+template <int D>
+int sumNodes(const FunctionTreeVector<D> & fs) {
```
stigrj: Please use `sum_nodes()` for free functions, and `sumNodes()` for class functions

src/trees/FunctionTreeVector.h
```diff
@@ -1,38 +1,53 @@
...
+      nNodes += f->getNNodes();
+    }
+  }
+  return nNodes;
+}
 
-    FunctionTree<D> *operator[](int i);
-    const FunctionTree<D> *operator[](int i) const;
+template <int D>
+double getCoef(const FunctionTreeVector<D> & fs, int i) {
```
stigrj: `get_coef()`

src/trees/FunctionTreeVector.h
```diff
@@ -1,38 +1,53 @@
...
+  return std::get<0>(fs[i]);
+}
 
-protected:
-    std::vector<double> coefs;
-    std::vector<FunctionTree<D> *> funcs;
-};
+template <int D>
+FunctionTree<D> & getFunc(FunctionTreeVector<D> & fs, int i) {
```
stigrj: `get_func()`

tests/operators/helmholtz_operator.cpp
```diff
@@ -72,7 +72,7 @@ TEST_CASE("Helmholtz' kernel", "[init_helmholtz], [helmholtz_operator], [mw_oper
 
                 MWOperator O(oper_mra);
                 for (int i = 0; i < K.size(); i++) {
-                    FunctionTree<1> &kern_tree = *K[i];
+                  FunctionTree<1> &kern_tree = getFunc(K, i);
```
stigrj: Indentation

src/operators/ConvolutionOperator.cpp
stigrj: Aha, this is by necessity, not by choice?

src/treebuilders/multiply.h
robertodr: I couldn't compile otherwise, but I forgot to try harder after getting it to work.

src/operators/ConvolutionOperator.cpp
robertodr: `push_back` for `std::vector<T>` will only accept stuff of type `T` as arguments. The alternative is to have a free function: 
```
template <int D>
void push_back(FunctionVectorTree<D> &fs, const FuntionTree<D> * f, double coef = 1.0);
```
but then it looks a bit weird (I think) at call site: 
```
push_back(tree_vec, k_tree);
```
I have been looking at overloading `push_back` from `std::vector<T>` and it seems to be explicitly forbidden by the standard...

src/trees/FunctionTreeVector.h
robertodr: Is this convention documented? :)

tests/operators/helmholtz_operator.cpp
robertodr: Yup, the editor is all over the place.

src/trees/FunctionTreeVector.h
robertodr: BTW, just noticed `sumNodes` (or the prospective `sum_nodes`) is not used anywhere within MRCPP. Is it used in MRChem or can it be removed altogether?

src/trees/FunctionTreeVector.h
robertodr: Done.

src/trees/FunctionTreeVector.h
robertodr: Done.

src/trees/FunctionTreeVector.h
robertodr: Done.

src/treebuilders/multiply.h
robertodr: Done, partially. `FunctionTreeVector` is defined as a template alias and it does not look like this construct plays well with forward declarations: https://stackoverflow.com/questions/16304468/forward-declare-a-template-alias

stigrj:
I think you forgot to `git rm FunctionTreeVector.cpp`

src/operators/ConvolutionOperator.cpp
stigrj: Hmm, `CoefsFunctionTree<D>` has a constructor that takes `(double, FunctionTree<D> *)`. It's weird that this constructor is not picked up automatically if you call `push_back` with these two arguments (I've tried). Probably because of the template...

src/trees/FunctionTreeVector.h
stigrj: Yes, it's used in MRChem and is sort of part of the API

stigrj: And the naming conventions will be suddenly documented :wink:

stigrj:
Greatness! Beware, I squashed it.

PR #34, opened 2018-05-15
Use list initialization rather than std::make_tuple

robertodr:
A bit less verbose follow-up to #33. C++11 lets you do:
- either `tree_vec.push_back(std::make_tuple(coef, func));`
- or `tree_vec.push_back({coef, func});` (list initialization)
I prefer the latter, but both are legal and have the same effect. I have changed the code in MRCPP to use the list initialization.


robertodr:
Dang! It only works with a relatively recent standard C++ library! See here: https://stackoverflow.com/a/14961849

PR #35, opened 2018-05-23
Vector operations

stigrj:
Imported the following vector operations from mrchem:
- FunctionTreeVector = gradient(FunctionTree)
- FunctionTree = divergence(FunctionTreeVector)
- FunctionTree = dot(FunctionTreeVector, FunctionTreeVector)

Also had to move the following `dot` to the same file as the above `dot` because of reasons (but it probably belongs there anyway):
- double = dot(FunctionTree, FunctionTree)

### DIFF
--- a/src/treebuilders/add.cpp
+++ b/src/treebuilders/add.cpp
@@ -12,17 +12,61 @@
 
 namespace mrcpp {
 
+/** @brief Addition of two MW function representations
+ *
+ * @param[in] prec Build precision of output function
+ * @param[in,out] out Output function to be built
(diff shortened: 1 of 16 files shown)


stigrj:
Documented all API tree builders. Even threw in some tests :slightly_smiling_face:

stigrj:
Thanks! Unfortunately, we cannot use it in MRChem at the moment because MRCPP/master is not compatible with MRChem. Can't remember exactly what the problem was, but I believe it was CMake related and that the [super build](https://github.com/MRChemSoft/mrchem/pull/96) would solve it.

PR #41, opened 2018-07-05
Bugfix: receiving cropped trees

stigrj:
Bugfix
-------
- The function `getNChunksUsed()` returned the actual number of used chunks, when it ought to return the position of the *last* used chunk. If entire chunks in the middle turned out empty after a `tree.crop()` they still needs to be counted if a node is present in the *last* chunk.

Changes
-----------
- Previously the `serialIx` was set to -1 just before `mpi::send_tree()` and `saveTree()`, now all serial indices (serial, parentSerial and childSerial) are cleared when chunks are first allocated *and* when nodes are deallocated.


stigrj:
Not when there are holes in the chunks

gitpeterwind:
SerialFunctionTree->nNodes is supposed to give the last one, also if there are holes

PR #44, opened 2018-09-10
Nullptrs and namespace

bjorgve:
I have added some nullptrs and fixed maybe the most annoying 'feature' in our code, namely I replaced '//namespace' with '// namespace'.


stigrj:
Thanks! Note that I squashed your commits, so you need to start fresh from `upstream/master`.

PR #45, opened 2018-09-22
shared memory for rcv tree and load tree

gitpeterwind:
The shared memory was not used when extending trees sent with mpi.
Also included when into LoadTree (but not tested)


stigrj:
Is this related to the weird behavior of the shared density?

gitpeterwind:
Yes: when trees grow, the memory was not allocated in the shared pool, if the growing happens while receiving a tree. There was a logical explanation after all :)

PR #46, opened 2018-09-25
[WIP] STL_containers in constructors

bjorgve:
Making several constructors accept std::vector<type> as input. This makes implementing vampyr functionality much easier. As a bonus it's a step towards a modern code.

### DIFF
--- a/doc/mrcpp_api/introduction.rst
+++ b/doc/mrcpp_api/introduction.rst
@@ -77,7 +77,7 @@ which can be written as the lambda function
     auto f = [Z] (const double *r) -> double {
         double R = sqrt(r[0]*r[0] + r[1]*r[1] + r[2]*r[2]);
         return Z/R;
-    }
+    };
 
 Note that the function signature must be *exactly* as given above, which means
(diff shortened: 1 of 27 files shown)


stigrj:
You'll need to rebase on `upstream/master`

robertodr:
Nice. Will be especially useful to translate `double * r` to `std:array<double, D> r`. Is there some underlying design issue to having the data member themselves be STL containers? Here you pass `std::vector` but then "unpack" the raw pointer using `.data()`.

stigrj:
The data is passed down to the `NodeIndex` class which was originally intended to be as simple and lightweight as possible, since it will be used millions and billions of times in a full calculation. Also we explicitly avoided any non-fundamental data types at this low level to simplify MPI routines.

bjorgve:
What's the advantage of using `std::array<double, D>` over 'std::vector<double>` ? In vampyr both lists and np.arrrays are accepted as input when `std::vector` is used, but with `std::array` only regular lists are accepted.

bjorgve:
I'll go for `std::array`. I don't think @Gabrielgerez will mind changing his `np.array` to regular lists.

PR #48, opened 2018-10-04
Read-only permissions for PHOpeator coefficients

stigrj:
Need to fix on the MRChem side as well


robertodr:
Was this tested to fix the problem?

ilfreddy:
@robertodr: not yet. But I "reverse-tested" it by granting write permission to the files and the code works again.
@stigrj: any other places where this happens? It seems to me that the filter matrices and the CCC are accessed read-only.
Otherwise good to go!

stigrj:
Yes, the other matrices in MRCPP are read-only, but we have the same issue in MRChem with the GTO basis set

PR #50, opened 2018-10-12
Introduce Coord<D> type

stigrj:
- Define `mrcpp::Coord<D> = std::array<double, D>`
- Replace all `evalf` arguments
- Replace macro `define MWNodeVector` with `using MWNodeVector`

### DIFF
--- a/examples/derivative.cpp
+++ b/examples/derivative.cpp
@@ -3,8 +3,6 @@
 #include "MRCPP/Printer"
 #include "MRCPP/Timer"
 
-#include <array>
-
 const auto min_scale = -4;
 const auto max_depth = 25;
 
@@ -34,11 +32,11 @@ int main(int argc, char **argv) {
     // Setting up analytic functions
     const auto alpha = 3.0;
     const auto r_0 = mrcpp::pi - 3.0;
-    auto f = [alpha, r_0] (const double *r) -> double {
+    auto f = [alpha, r_0] (const mrcpp::Coord<D> &r) -> double {
         auto R = fabs(r[0] - r_0);
         return exp(-alpha*R);
     };
-    auto df = [alpha, r_0] (const double *r) -> double {
+    auto df = [alpha, r_0] (const mrcpp::Coord<D> &r) -> double {
         auto R = fabs(r[0] - r_0);
         auto sign = 1.0;
         if (r[0] > r_0) sign = -1.0;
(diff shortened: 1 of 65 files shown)


examples/derivative.cpp
```diff
@@ -34,11 +32,11 @@ int main(int argc, char **argv) {
...
     const auto alpha = 3.0;
     const auto r_0 = mrcpp::pi - 3.0;
-    auto f = [alpha, r_0] (const double *r) -> double {
+    auto f = [alpha, r_0] (const mrcpp::Coord<D> &r) -> double {
         auto R = fabs(r[0] - r_0);
         return exp(-alpha*R);
     };
-    auto df = [alpha, r_0] (const double *r) -> double {
+    auto df = [alpha, r_0] (const mrcpp::Coord<D> &r) -> double {
         auto R = fabs(r[0] - r_0);
```
robertodr: Nitpick, but replacing `fabs` with `std:abs` will work for any numerical type, not only floats/

src/functions/GaussFunc.cpp
```diff
@@ -324,10 +319,10 @@ double GaussFunc<3>::calcCoulombEnergy(GaussFunc<3> &gf) {
 
     BoysFunction boys(0);
 
-    double boysArg = alpha*Rpq_2;
-    double boysFac = boys.evalf(&boysArg);
+    Coord<1> boysArg = {alpha*Rpq_2};
```
robertodr: Is the `=` needed here when using brace initialization?

tests/functions/polynomial.cpp
```diff
@@ -55,15 +55,15 @@ TEST_CASE("Polynomial evaluation", "[poly_evalf], [polynomials]") {
     Polynomial P(c, &a, &b);
 
     SECTION("Evaluation within bounds") {
-        double x = 1.5;
-        double calc_val = P.evalf(&x);
-        double ref_val = 1.0 + 2.0*x*x;
+        Coord<1> x = {1.5};
```
robertodr: I think `Coord<1> x{1.5};` might achieve the same effect.

examples/derivative.cpp
stigrj: Yes, I'm replacing them whenever I see them. I guess I missed this one

tests/functions/polynomial.cpp
stigrj: Hmm, I found this version better than `x({1.5})` which I had first, but if it works without the `()`, then I agree

robertodr: It should work _without_ `()` there are some subtleties for copying _vs_ initializing, but it shouldn't matter at all in this case.

stigrj:
I think it's pretty :unicorn:

robertodr:
Me too :unicorn: Also, it makes clearer what this _means_ rather than how it's implemented, which might help newcomers who know what a coordinate vector is, but not necessarily what a `std::array<double, D>` is.

PR #58, opened 2018-11-26
std::array and Coord<D> in constructor of GaussPoly

bjorgve:
Need GaussPoly in vampyr so made GaussPoly constructors with std::array and Coord<D>.

I also updated the docs with an example of how to use it in mrcpp


bjorgve:
@stigrj turns out I'm still not able to import the GaussPoly class, so maybe wait a bit with approving. I am not sure if further fixing has to appear on the mrcpp or vampyr side of the code.

PR #59, opened 2018-12-05
[WIP] Faster, better, stronger Gaussians

bjorgve:
Adding gaussians with variable exponents in x, y and z directions. 

Such as these:

<a href="https://www.codecogs.com/eqnedit.php?latex=\exp(\alpha_x&space;x^2&space;&plus;&space;\alpha_z&space;z^2&space;&plus;&space;\alpha_z&space;z^2)" target="_blank"></a>

Currently (in practice)  this works for GaussFunc, with non or few edits to the sutler parts of 
the implementation.  

Happy for inputs on places I need to edit, there probably are things I don't think about.

- [x]  Project onto f_trees
- [x] Finish Gaussian
- [x] Finish GaussFunc
- [x] Finish GaussPoly
- [x] Update unit texts


cmake/compiler_flags/GNU.CXX.cmake
```diff
@@ -1,6 +1,6 @@
 if(NOT DEFINED ENV{CXXFLAGS})
     if(CMAKE_CXX_COMPILER_ID MATCHES GNU)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-unknown-pragmas -Wno-sign-compare -Woverloaded-virtual -Wwrite-strings -Wno-unused")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wno-unknown-pragmas -Wno-sign-compare -Woverloaded-virtual -Wwrite-strings -Wno-unused")
```
robertodr: C++14! Nice! I'd suggest to set this globally as a CMake property:
```
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
set(CMAKE_CXX_EXTENSIONS FALSE)
```
CMake will then be able to pick up later standards, it the compiler supports them.

bjorgve: That would be awesome, where do I add it? The CMakeLists.txt? (It tells me not to edit it.)

robertodr: I would add a top-level file `CXXFlags.cmake` in the `compilers` folder with these contents:
```
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
set(CMAKE_CXX_EXTENSIONS FALSE)
set(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)

include(GNU.CXX)
include(Intel.CXX)
````
You could do the same for the C flags (is there actually any C in MRCPP?) and set the standard. You'd need to modify `autocmake.yml` accordingly and run `python update.py ..` to get the changes propagated.

In XCFun we put in place something a little bit more elaborate where the validity of the flags is actually checked: https://github.com/dftlibs/xcfun/tree/master/cmake/custom/compilers but this is a bit off-topic, of course. Ping me if you need help, of course.

Comment (bjorgve) [reply cmake/compiler_flags/GNU.CXX.cmake]: Okei! I'll see if I can figure it out tomorrow! Not sure, probably not. Do you know Stig? @stigrj

Comment (stigrj) [reply cmake/compiler_flags/GNU.CXX.cmake]: Have a look at the vampyr setup, I think it looks like the one @robertodr is suggesting. We should do the same in MRChem, by the way. I also think the C flags can be removed.

Comment (robertodr) [reply cmake/compiler_flags/GNU.CXX.cmake]: Yes, @stigrj is right: https://github.com/MRChemSoft/vampyr/blob/master/cmake/compiler_flags/CXXFlags.cmake

Comment (bjorgve) [reply cmake/compiler_flags/GNU.CXX.cmake]: Think I fixed it.

Comment (robertodr) [reply cmake/compiler_flags/GNU.CXX.cmake]: There's two `CXXFlags.cmake` files now :thinking: and the `-std=c++14` flag is still in the list.

Comment (bjorgve) [reply cmake/compiler_flags/GNU.CXX.cmake]: Removed the flags. You see two in cmake/compiler_fags?

Comment (robertodr) [reply cmake/compiler_flags/GNU.CXX.cmake]: I see this https://github.com/MRChemSoft/mrcpp/pull/59/files#diff-528afcc391b811de7797116996511692 and this https://github.com/MRChemSoft/mrcpp/pull/59/files#diff-b191b6d54789eaf48c5365cb022ca5b4

Comment (bjorgve) [reply cmake/compiler_flags/GNU.CXX.cmake]: Okei, how do I fix it?

File: src/functions/GaussFunc.cpp [hunk 2]
```diff
@@ -75,7 +77,7 @@ double GaussFunc<D>::evalf(double r, int d) const {
     } else {
         p2 = std::pow(q, this->power[d]);
     }
-    double result = p2 * std::exp(-this->alpha * q2);
+    double result = p2 * std::exp(-this->alpha[0] * q2);
```
stigrj: Should it not be `alpha[d]`?

src/functions/GaussFunc.cpp
```diff
@@ -84,7 +86,7 @@ template<int D>
 double GaussFunc<D>::calcSquareNorm() {
     double norm = 1.0;
     for (int d = 0; d < D; d++) {
-        double a = 2.0 * this->alpha;
+        double a = 2.0 * this->alpha[0];
```
stigrj: `alpha[d]`

cmake/compiler_flags/GNU.CXX.cmake
bjorgve: Ah ofc sorry! For some reason I thought they both where in the same directory, and I could only see one.

PR #60, opened 2018-12-11
Even faster, better and stronger gaussians

bjorgve:


stigrj:
I think there might be some error in `.travis.yml` that prevents Travis from running. If you go to "More options -> Requests" in the top right-hand corner on Travis you find the error message "Could not parse .travis.yml" in the last 5 commits.

bjorgve:
I reverted travis to the state given in mrchemsoft/master. Travis now runs, but it crashes when in tries to install cmake. Do you know why @robertodr?  

`home/travis/.travis/job_stages: line 104: ./.ci/cmake.sh: Permission denied`
`The command "./.ci/cmake.sh" failed and exited with 126 during .`

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: cpp
 sudo: false
+dist: trusty
```
robertodr: Travis has introduced Xenial in the past few weeks. I suggest using it with: `dist: xenial`.

src/functions/GaussExp.h
```diff
@@ -66,7 +66,7 @@ class GaussExp: public RepresentableFunction<D> {
     void operator*=(double d) { this->multInPlace(d); }
 
     double getScreening() const { return screening; }
-    double getExp(int i) const { return this->funcs[i]->getExp(); }
+    auto getExp(int i) const { return this->funcs[i]->getExp(); }
```
robertodr: Mmmm... I'm not so sure about return type deduction. This makes is pretty hard to understand what's going on for people reading the code, unless they use an IDE (which I guess is _not_ the majority of people working on MRCPP and MRChem) It might come in handy for templated functions, where it's hard or impossible to use the trailing return type declaration...

src/functions/GaussFunc.cpp
```diff
@@ -279,7 +287,17 @@ double GaussFunc<D>::calcCoulombEnergy(GaussFunc<D> &gf) {
 
 template<int D>
 std::ostream& GaussFunc<D>::print(std::ostream &o) const {
-    o << "Exp:   " << this->getExp() << std::endl;
+    auto expTmp = this->getExp();
+    auto is_array = std::all_of(expTmp.begin(), expTmp.end(),
```
robertodr: Is this concoction checking whether there are elements beyond `expTmp[0]`? It's a bit obfuscated, I think. Since I can't figure out a suggestion for a cleaner way to do it, I'll just say insert a comment here.

The lambda captures `expTmp` by value, but probably capturing `std::begin(expTmp)` is enough.

src/functions/GaussFunc.cpp
```diff
@@ -298,8 +316,23 @@ std::ostream& GaussFunc<D>::print(std::ostream &o) const {
  */
 template<>
 double GaussFunc<3>::calcCoulombEnergy(GaussFunc<3> &gf) {
-    double p = this->getExp();
-    double q = gf.getExp();
+
+    auto p_tmp = this->getExp();
+    auto is_all_same = std::all_of(p_tmp.begin(), p_tmp.end(),
```
robertodr: Given that this `all_of` pattern repeates, I suggest you factor it out into its own little function in a `detail` namespace.

src/functions/GaussFunc.cpp
```diff
@@ -298,8 +316,23 @@ std::ostream& GaussFunc<D>::print(std::ostream &o) const {
...
+                                [p_tmp](double i) {return i == *p_tmp.begin(); });
+    if (!is_all_same)
+        NOT_IMPLEMENTED_ABORT;
+
+    auto p = p_tmp[0];
+    auto q_tmp = gf.getExp();
+    is_all_same = std::all_of(q_tmp.begin(), q_tmp.end(),
+                                [q_tmp](double i) {return i == *q_tmp.begin(); });
```
robertodr: Same comment on the lambda capture.

src/functions/GaussPoly.cpp
```diff
@@ -249,7 +294,17 @@ void GaussPoly<D>::setPoly(int d, Polynomial &poly) {
 
 template<int D>
 std::ostream& GaussPoly<D>::print(std::ostream &o) const {
-    o << "Exp: " << this->getExp() << std::endl;
+    auto expTmp = this->getExp();
+    auto is_array = std::all_of(expTmp.begin(), expTmp.end(),
```
robertodr: Ditto on factoring out `all_of` pattern and the lambda capture.

src/functions/Gaussian.cpp
```diff
@@ -16,7 +16,7 @@ namespace mrcpp {
 
 template<int D>
 Gaussian<D>::Gaussian(double a, double c, const double r[D], const int p[D]) {
```
robertodr: I'd use a CTOR list, rather than initializing in the body.

.ci/cmake.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
robertodr: You should make these two files executable for Travis not to choke on them. That is: `chmod u+x cmake.sh`

src/functions/GaussExp.h
stigrj: agreed

src/functions/Gaussian.cpp
bjorgve: I'm not able to use CTOR on `const double r[D]` and `const int p[D]` (I think). This constructor should probably be removed, but I am keeping it to avoid breaking any client code.

src/functions/GaussFunc.cpp
robertodr: Maybe use [`std::equal`](https://en.cppreference.com/w/cpp/algorithm/equal)?

bjorgve: I was not able to make it check if all in the array are the same. Do you have a code example?

robertodr: Actually, after trying it out it doesn't work as I had expected :sweat_smile:

stigrj:
@bjorgve do you consider this merge-able? Or did you still have some CMake trouble on Travis?

bjorgve:
I consider it merge-able, although I think it should be better tested. But I don't know what to test and how thorough the tests should me.

PR #61, opened 2018-12-17
routine to print out memory usage

gitpeterwind:
Should give total memory usage for the process calling.


src/utils/Printer.cpp
```diff
@@ -154,4 +147,35 @@ int Printer::setPrecision(int i) {
...
+    char* s = p;
+    while (*s >='0' && *s <= '9'){ s++; line[len] = p[len]; len++;}
+    p[len] = 0;
+    return atoi(p);
+}
+
+/** Prints (and returns) the current memory usage of this process
+ */
+int Printer::printMem(char* txt){
+    FILE* file = fopen("/proc/self/statm", "r");
```
stigrj: Is this system independent?

gitpeterwind: I do not think so. Or maybe it must be linux. I must say I do not understand the details: it is not a real file, as it might look like!
"The proc filesystem is a pseudo-filesystem which provides an
       interface to kernel data structures.  It is commonly mounted at /proc. "

gitpeterwind: Wrong/ambiguous answer before: I meant "I think this is system independent!"

PR #64, opened 2019-01-09
[WIP] Simple periodic

bjorgve:
Rebased the periodic implementation from last year onto current mrcpp. Make it compile, all tests runs fine aprart from the convolution operators (SEGFAULT). I will look into it tomorrow, @stigrj you see why easily? 

I will rebase it onto mrcpp when the scaling function is merged.


bjorgve:
Wait with commenting on this. The rebase turned out really messy so I am looking into adding things manually.

PR #67, opened 2019-01-16
B-spine derivative operator

bjorgve:
- [x] Ready for review

Make it work for:
- [x] Legendre scaling function
- [x] Interpolating scaling function

It seems to work well now both for Legendre and for Interpolating basis functions. 

As a side note. The derivatives should probably be tested for all polynomial orders,
not a single one as it is today.


src/treebuilders/BSCalculator.cpp
```diff
@@ -0,0 +1,126 @@
...
+        case -1:
+            for (int i = 0; i < kp1; i++) {
+                for (int j = 0; j < kp1; j++) {
+                    int idx = i * kp1 + j;
+                    coefs[2 * kp1_d + idx] = two_np1 * this->S_m1(i, j);
+                }
+            }
+            break;
+        default:
+            MSG_ERROR("This translation should not occour");
```
stigrj: I think this warning had to be removed with PBC?

bjorgve: Yepp! Fixed

PR #73, opened 2019-01-23
Update Autocmake

robertodr:
Resolve a couple of issues when using MRCPP as an external dependency:

1. `CMAKE_MODULE_PATH` might have been populated by the superproject and include CMake modules that conflict with those in MRCPP. This was solved in https://github.com/dev-cafe/autocmake/pull/256
2. Use `PROJECT_{BINARY,SOURCE}_DIR` rather than `CMAKE_{BINARY,SOURCE}_DIR`. The former is within the scope of MRCPP as a subproject, whereas the latter will refer to the superproject.


robertodr:
Similar update of Autocmake might be needed in MRChem. I am looking at VAMPyR myself now, as I wanted to create a Docker image to run it interactively on MyBinder

bjorgve:
It doesn't seem to work for me.

stigrj:
The compiler flag CMake files are not found: `cmake/compiler_flags/GNU.CXX.cmake` etc.

robertodr:
Yeah. I fixed it, but forgot to push.

PR #74, opened 2019-01-25
Cropchunk

gitpeterwind:
Call to "shrinkChunks" after "crop". This actively moves nodes from the tree to fill holes lower down on the stack. The unused chunks are then deleted. Add in place (used for example in rotation) was adding many holes, and a lot of memory was lost. The more MPI the more holes; a test show that with 8 MPI already about 30% memory is saved with the new version. 
Numerical results should not be affected at all (to the last digit).
Other small changes: shared memory with a rank in the class; printMem can be made silent.


src/trees/SerialFunctionTree.cpp
```diff
@@ -479,6 +580,7 @@ template <int D> void SerialFunctionTree<D>::rewritePointers(int nChunks) {
                     int n_ichunk = node->parentSerialIx / this->maxNodesPerChunk;
                     int n_inode = node->parentSerialIx % this->maxNodesPerChunk;
                     node->parent = this->nodeChunks[n_ichunk] + n_inode;
+                    assert(node->parent->serialIx != node->parentSerialIx);
```
stigrj: This assertion fails in some unit tests. Is it not supposed to be `==`?

gitpeterwind: Yes, you are right!
Where can I see where it went wrong? (I tried to get to "details", but did not see this part.

stigrj: First "details", then click on the failing build and scroll to the bottom. There's usually a LOT of scrolling because of verbose build.

PR #75, opened 2019-01-25
CMake refactoring

robertodr:
I mostly wanted to be able to fetch Eigen3 if it is not available on the machine, but it sort of snowballed from there:

- Fetch Eigen3 if not locally available
- Built either shared or static. Building both correctly would require building sources twice. This can be controlled with the flag `BUILD_STATIC_LIBS`.
- Rework installation of headers and definiton of include directories. It is slightly simpler and leads to more compact `CMakeLists.txt`
- Removed the include folder. Those headers are now under api and tucked under `${CMAKE_INSTALL_INCLUDEDIR}/MRCPP`.

~**UPDATE**
I am testing this in the wild with MRChem. Still a bit rough around the edges.~


robertodr:
You probably want to squash-merge this one.

bjorgve:
Cool! I'll look at it first thing in the morning!

stigrj:
It's when building the `mrcpp-tests` target. Without `--enable-tests` it works.

stigrj:
Hmm, perhaps this goes also for the real applications, that the collective headers like `MRCPP/MWFunctions` won't work with Clang?

robertodr:
I'll try with Clang locally and see what's wrong. I can't see why API headers shouldn't work, it could be some subtle issue with symbol visibility. I am never quite sure with templates...

stigrj:
This fixed the error locally on my machine, but not on Travis it seems

robertodr:
I think the template functions used for testing should go into an header file, so that there are no ambiguities at the linking stage.

bjorgve:
Things seems to be working fine at my end

robertodr:
Clang or GCC?

bjorgve:
GCC, running Clang now. Are the warnings new? Should we may make be pose an issue on it?

robertodr:
@bjorgve the warnings seem to arise from `-Weverything` and I think should be fixed, so feel free to post an issue. Seems to be a tedious, but straightforward exercise for an hackathon session. @stigrj the infamous `-Ofast` is still among the flags for MRCPP. Should I nuke it in favor of `-O2` or `-O3`?

robertodr:
Disgusting, now the superbuild PR fails... Need to think a bit more about this.

stigrj:
Yes to the `-Ofast` nukleation. In MRChem we keep `-O3` for GNU, but `-O2` for Intel, so you can do the same here.

robertodr:
@stigrj I managed to build and test on fram with `module load Boost/1.68.0-intel-2018b-Python-3.6.6 CMake/3.12.1`. Can you double-check the latest commit fixes your issues too?

stigrj:
For some reason the `~Gaussian<D> = default;` was located in the source file, so the Intel linker couldn't find it. Did you compile with Intel `--cxx=icpc`? Now the build and tests work, but the Eigen fetch still fails with the following error:
```
-- Configuring incomplete, errors occurred!
See also "/cluster/home/stig/src/mrcpp-stig/build-super/CMakeFiles/CMakeOutput.log".
See also "/cluster/home/stig/src/mrcpp-stig/build-super/CMakeFiles/CMakeError.log".

CMake Error at /cluster/home/stig/src/mrchem-super/build-intel/_deps/eigen3_sources-build/Eigen3Config.cmake:20 (include):
  include could not find load file:

    /cluster/home/stig/src/mrchem-super/build-intel/_deps/eigen3_sources-build/Eigen3Targets.cmake
Call Stack (most recent call first):
  external/upstream/fetch_eigen3.cmake:3 (find_package)
  cmake/custom/main.cmake:15 (include)
  CMakeLists.txt:64 (include)
```

I don't understand why it looks in the `/mrchem-super/` directory, though, I'm currently building `/mrcpp-super/`

stigrj:
Seems it looks in any other folder where I have previously attempted an Eigen super-fetch. Still fails with the same error if I remove all those, only now with the current directory path.

robertodr:
Yes, I am building with `icpc` and I cannot reproduce neither the DTOR nor the Eigen failures. What's in your `PATH`? And which Intel module are you loading?

robertodr:
Can reproduce the DTOR failure with `module load Boost/1.63.0-intel-2017a-Python-2.7.13 CMake/3.12.1`. Later versions of Intel seem to be more lenient then...

robertodr:
@ilfreddy @bjorgve @gitpeterwind @bast can anyone else of you reproduce @stigrj's problems on fram?

stigrj:
Ok, now it works for MRCPP (can't see that I did anything different from before...), but still not for MRChem

robertodr:
What's the problem with mrchem? Still Eigen?

stigrj:
Yes, the same Eigen error. I've tried a dozen different approaches

robertodr:
I'm at a loss, I must admit. I suggest moving further discussion to MRChemSoft/mrchem#198, if the issue for standalone MRCPP has vanished/was solved.

PR #76, opened 2019-01-27
silent printMem

gitpeterwind:


src/utils/Printer.cpp
```diff
@@ -187,7 +187,7 @@ int Printer::printMem(char *txt, bool silent) {
     char line[80];
     while (fgets(line, 80, file) != NULL) {
```
bjorgve: Maybe change NULL to nullptr when you're here

src/utils/Printer.cpp
```diff
@@ -187,7 +187,7 @@ int Printer::printMem(char *txt, bool silent) {
...
         val = getVal(line, 6); //sixth number is data+stack in pages (4kB)
-        printf("Mem usage %s %6.3f %s \n", txt, val * 4.0 / (1024.0 * 1024), "GB");
+        if (not silent) printf("Mem usage %s %6.3f %s \n", txt, val * 4.0 / (1024.0 * 1024), "GB");
```
bjorgve: It's probably better with a std::cout

PR #82, opened 2019-02-04
Fix wart in INTERFACE_INCLUDE_DIRECTORIES

robertodr:
This should fix @stigrj problems in superbuilding MRChem from a preinstalled MRCPP.


stigrj:
Now it fails for the API headers which already have a `MRCPP/` prefix...

It works if I use both, like we do in the `BUILD_INTERFACE`:
```
$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}>         
$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/MRCPP>   
$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>                             
$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/MRCPP> 
```
Perhaps we should add a `MRCPP/` prefix on all internal includes in MRCPP? Then we probably get away with
```
$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}>         
$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>                             
```

stigrj:
What kind of forgiving system are you working on, where everything just works? :stuck_out_tongue:

stigrj:
Ok, if I instead use the following:
```
INTERFACE                                                                      
   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}>         
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
```
and build MRCPP I get a "redefinition of class" error on `class RepresentableFunction` and `class Polynomial`, but if I switch to the old school include guards
```
#ifndef POLYNOMIAL_H_
#define POLYNOMIAL_H_

#endif // POLYNOMIAL_H_
```
it suddenly works all the way up to MRChem!

stigrj:
Not sure what to do with the include guards. Switch back to `#ifndef`s in the whole project?

stigrj:
Fixed in #83

PR #84, opened 2019-02-11
new better definition of getNchunks that works for overwritten trees

gitpeterwind:
New defintion of getNChunksUsed. The old one counted all occupied nodes, even if they were not part of the tree. This did not necessarily work for overwritten trees, as they may have nodes closer to the end, that were not overwritten (if the tree had more chunks than what is overwritten).
The new formulation is also simpler and faster.

Also somewhat safer rewritePointers. It would be even better to rewrite the routine, so that it traverse the tree, instead of all chunks; that would clean the trees of possible orphan nodes. Also the "resetendnodetable" could be integrated here.


gitpeterwind:
I will try to make a new rewritepointers before I resubmit

PR #86, opened 2019-02-12
Serialtrees

gitpeterwind:
New defintion of getNChunksUsed. The old one counted all occupied nodes, even if they were not part of the tree. This did not necessarily work for overwritten trees, as they may have nodes closer to the end, that were not overwritten (if the tree had more chunks than what is overwritten).
The new formulation is also simpler and faster.

RewritePointers is reformulated so that it traverse the tree instead of going through all the chunks. This will then avoid possible problems with orphan nodes, since those will not be counted/included.
Also the call to resetEndNodeTable is now removed, by including it directly. Also simpler and faster.

### DIFF
--- a/src/trees/SerialFunctionTree.cpp
+++ b/src/trees/SerialFunctionTree.cpp
@@ -569,8 +569,8 @@ template <int D> void SerialFunctionTree<D>::rewritePointers(int nChunks) {
             ProjectedNode<D> *node = (this->nodeChunks[ichunk]) + inode;
             if (node->serialIx >= 0) {
                 this->nNodes = ichunk * this->maxNodesPerChunk + (inode + 1);
-
                 // Node is part of tree, should be processed
+                assert(node->serialIx == this->nNodes - 1);
                 this->getTree()->incrementNodeCount(node->getScale());
                 if (node->isEndNode()) this->getTree()->squareNorm += node->getSquareNorm();
 
@@ -599,6 +599,8 @@ template <int D> void SerialFunctionTree<D>::rewritePointers(int nChunks) {
                     node->children[i] = this->nodeChunks[n_ichunk] + n_inode;
                 }
                 this->nodeStackStatus[node->serialIx] = 1; // occupied
+            } else {
+                this->nodeStackStatus[node->serialIx] = 0; // available
             }
         }
         this->lastNode = this->nodeChunks[ichunk] + this->nNodes % (this->maxNodesPerChunk);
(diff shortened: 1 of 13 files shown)


gitpeterwind:
How do I "rebase" this?

stigrj:
To rebase on upstream master
```
git fetch upstream
git rebase upstream/master
```
Then you need to resolve the conflicts along the way, followed by
```
git add <files-with-conflicts>
git rebase --continue
```
Iterate until the rebase is completed. If you mess something up you can go back to the un-merged version
```
git rebase --abort
```
When you are done you need to overwrite the remote repo with a forced push
```
git push -f origin serialtrees
```

stigrj:
Actualy, in this case you can clean up bit by squashing some of the commits. If you do
```
git rebase -i upstream/master
```
you get into a git commit message. Here you can pick which of the commits you want to keep in the rebase. In this case we don't want the merge commits, so you can just remove those lines. Also it would be nice to squash the style commits together with the "real" commits. This is done by changing the `pick` keyword into `fixup`. This will squash the `fixup` commit into the previous one.

gitpeterwind:
Thanks! (Still don't feel very safe about what command changes the code or only the messages)

stigrj:
Great, look at the nice, clean history :slightly_smiling_face: Most of the tests fail, though. Do you know what's going on?

gitpeterwind:
The way to count nodes is different now, and one test failed. The tree clear was not (never) updating the serialtree counters; I tried to fix this by definining a clear for the serialtree when the tree is cleared. But that is giving a lot of segfaults apparently. Maybe the serial tree is not always defined if it is empty or so. Have to check, but Stallo is down...

stigrj:
Looks like it's `build_grid` that fails. E.i. allocating nodes but not coefficients.

src/trees/SerialFunctionTree.cpp
```diff
@@ -562,72 +569,65 @@ template <int D> void SerialFunctionTree<D>::rewritePointers(int nChunks) {
...
+
+        // adress of parent and children must be corrected
+        // can be on a different chunks
+        if (node->parentSerialIx >= 0) {
+            int n_ichunk = node->parentSerialIx / this->maxNodesPerChunk;
+            int n_inode = node->parentSerialIx % this->maxNodesPerChunk;
+            node->parent = this->nodeChunks[n_ichunk] + n_inode;
+            assert(node->parent->serialIx == node->parentSerialIx);
+        } else {
+            node->parent = 0;
```
stigrj: `nullptr`

gitpeterwind: ok, I corrected a bunch of them at the same time

PR #91, opened 2019-04-04
Periodic gaussians

bjorgve:
Periodic Gaussians
    
    Gaussians can now be made semi-periodic around the
    unit-cell of a given period. When these are projected
    onto a function tree, they are true periodic functions
    where the integral of the Gaussian is conserved within
    the unit-cell.

To make the Gaussian semi-periodic use the method
`makePeriodic`. The same also works for `GaussExp`.

EDIT:

We can now specify periodicity arbitrarily in each direction, I need to add some 
tests, but preliminary results seem promising.

- [x] Ready for review

### DIFF
--- a/src/functions/function_utils.cpp
+++ b/src/functions/function_utils.cpp
@@ -39,28 +39,29 @@ namespace mrcpp {
  * @param[out] shared_ptr<GaussExp<D>> semi-periodic version of a Gaussian around a unit-cell
  * @param[in] A Gaussian function that is to be made semi-periodic
  * @param[in] The period of the unit cell
+ * @param[in] Number of standard diviations covered in each direction. Default 4.0
  *
- * The Gaussian function replicated in neighbooring cells, such that atleast
- * 99.99367% if the integral can be captured within the unit-cell.
+ * nStdDev = 1, 2, 3 and 4 ensures atleast 68.27%, 95.45%, 99.73% and 99.99% of the
+ * integral is conserved with respect to the integration limits.
+ *
  */
 template <int D>
 std::shared_ptr<GaussExp<D>> function_utils::make_gaussian_periodic(Gaussian<D> &inp,
-                                                                    const std::array<double, D> &period) {
+                                                                    const std::array<double, D> &period,
+                                                                    double nStdDev) {
     auto gauss_exp = std::make_shared<GaussExp<D>>();
     auto pos_vec = std::vector<Coord<D>>();
 
-    // Four standard deviation upp and down from the expectation value of the
-    // gaussian holds approximately 99.99367% of the integral.
-    auto four_std = 4.0 * inp.getMaximumStandardDiviation();
+    auto x_std = nStdDev * inp.getMaximumStandardDiviation();
 
     // This lambda function  calculates the number of neighbooring cells
-    // requred to keep atleast 99.99367% of the integral conserved in the
+    // requred to keep atleast x_stds of the integral conserved in the
     // unit-cell.
-    auto neighbooring_cells = [period, four_std](auto pos) {
+    auto neighbooring_cells = [period, x_std](auto pos) {
         auto needed_cells_vec = std::vector<int>();
         for (auto i = 0; i < D; i++) {
-            auto upper_bound = pos[i] + four_std;
-            auto lower_bound = pos[i] - four_std;
+            auto upper_bound = pos[i] + x_std;
+            auto lower_bound = pos[i] - x_std;
             // number of cells upp and down relative to the center of the Gaussian
             needed_cells_vec.push_back(std::ceil(upper_bound / period[i]));
         }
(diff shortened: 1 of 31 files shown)


src/functions/function_utils.cpp
```diff
@@ -0,0 +1,121 @@
...
+ */
+template <int D>
+std::shared_ptr<GaussExp<D>> function_utils::make_gaussian_periodic(Gaussian<D> &inp,
+                                                                    const std::array<double, D> &period) {
+    auto gauss_exp = std::make_shared<GaussExp<D>>();
+    auto pos_vec = std::vector<Coord<D>>();
+
+    // Four standard deviation upp and down from the expectation value of the
+    // gaussian holds approximately 99.99367% of the integral.
+    auto four_std = 4.0 * inp.getMaximumStandardDiviation();
```
stigrj: Perhaps add `nStdDev` to the argument list of this function? Makes it a bit less hard coded

src/functions/function_utils.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    inp.setPos(startpos); // The starting position is set to complement build_grid
+                          // ensuring the projection of Gaussians with high exponents
+    auto nr_cells_upp_and_down = neighbooring_cells(startpos);
+    for (auto d = 0; d < D; d++) { startpos[d] -= nr_cells_upp_and_down * period[d]; }
+
+    auto tmp_pos = startpos;
+    for (auto x = 0; x < 2 * nr_cells_upp_and_down + 1; x++) {
+        if (D == 2 or D == 3) {
```
stigrj: This hurt my eyes! Can you try to rewrite using a recursive lambda function?
`auto` does not work in this case, so you'll have to spell it out, like
```
    std::function<int(int)> factorial;
    factorial = [&factorial](int i) {
        if (i == 0 or i == 1) {
            return 1;
        } else {
            return i*factorial(i-1);
        }
    };
```

src/trees/FunctionTree.cpp
```diff
@@ -204,12 +205,7 @@ template <int D> double FunctionTree<D>::evalf(const Coord<D> &r) const {
 
     // The 1.0 appearing in the if tests comes from the period is
     // always 1.0 from the point of view of this function.
-    if (this->getRootBox().isPeriodic()) {
-        for (auto i = 0; i < D; i++) {
-            if (arg[i] > 1.0) arg[i] = std::fmod(arg[i], 1.0);
-            if (arg[i] < 0.0) arg[i] = std::fmod(arg[i], 1.0) + 1.0;
-        }
-    }
+    if (this->getRootBox().isPeriodic()) { periodic::coord_mainpulation<D>(arg, this->getRootBox().getPeriodic()); }
```
stigrj: typo `mainpulation`

src/trees/FunctionNode.cpp
```diff
@@ -47,10 +48,7 @@ template <int D> double FunctionNode<D>::evalf(Coord<D> r) {
     // The 1.0 appearing in the if tests comes from the period is always 1.0
     // from the point of view of this function.
     if (this->getMWTree().getRootBox().isPeriodic()) {
-        for (auto i = 0; i < D; i++) {
-            if (r[i] > 1.0) r[i] = std::fmod(r[i], 1.0);
-            if (r[i] < 0.0) r[i] = std::fmod(r[i], 1.0) + 1.0;
-        }
+        periodic::coord_mainpulation<D>(r, this->getMWTree().getRootBox().getPeriodic());
```
stigrj: typo `mainpulation`

src/utils/periodic_utils.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "periodic_utils.h"
+#include "trees/NodeIndex.h"
+
+#include <cmath>
+
+namespace mrcpp {
+namespace periodic {
+
+template <int D> void indx_manipulation(NodeIndex<D> &idx, std::array<bool, D> periodic) {
```
stigrj: `const &` on the `periodic`?

src/utils/periodic_utils.cpp
```diff
@@ -0,0 +1,65 @@
...
+        if (periodic[i]) {
+            translation[i] = idx.getTranslation(i);
+            if (translation[i] >= two_n) translation[i] = translation[i] % two_n;
+            if (translation[i] < 0) translation[i] = (translation[i] + 1) % two_n + two_n - 1;
+        }
+    }
+    idx.setTranslation(translation);
+}
+
+template <int D> void coord_mainpulation(Coord<D> &r, std::array<bool, D> periodic) {
```
stigrj: `const &periodic`

src/utils/periodic_utils.cpp
```diff
@@ -0,0 +1,65 @@
...
+    int two_n = 1 << idx.getScale();
+    for (auto i = 0; i < D; i++) {
+        if (periodic[i]) {
+            translation[i] = idx.getTranslation(i);
+            if (translation[i] >= two_n) translation[i] = translation[i] % two_n;
+            if (translation[i] < 0) translation[i] = (translation[i] + 1) % two_n + two_n - 1;
+        }
+    }
+    idx.setTranslation(translation);
+}
```
stigrj: `translation[i]` remains uninitialized if `periodic[i] == false`

stigrj: Also, the bit shift will not work if `scale` is negative. This should not happen in periodic calculations, but I think you can still add a guard for this.

bjorgve: Fixed :)

PR #92, opened 2019-06-21
Revamp Plotter class

stigrj:
See documentation changes for details


doc/mrcpp_api/plotter.rst
```diff
@@ -5,33 +5,46 @@ Plotter
...
+This class will generate an equidistant grid in one (line), two (surf)
+or three (cube) dimensions, and subsequently evaluate the function on
+this grid.
+
+The grid is generated from the vectors A, B and C in relation to the origin O:
+ - a linePlot will plot the line spanned by A, starting from O
+ - a surfPlot will plot the area spanned by A and B, starting from O
+ - a cubePlot will plot the volume spanned by A, B and C, starting from O
+
+The vectors A, B and C does not necessarily have to be orthogonal.
```
robertodr: Typo "does not" --> "do not"

src/utils/Plotter.cpp
```diff
@@ -100,236 +63,336 @@ template <int D> void Plotter<D>::setSuffix(int t, const std::string &s) {
...
+    Plots the function func parametrically along the vector A starting
+    from the origin O to a file named fname + file extension (".line"
+    as default).
 */
-template <int D> void Plotter<D>::surfPlot(const RepresentableFunction<D> &func, const std::string &fname) {
-    println(20, "--------Surface Plot----------");
-    if (not verifyRange()) MSG_ERROR("Zero range");
-    calcSurfCoordinates();
-    evaluateFunction(func);
+template <int D>
```
robertodr: I think you can unify the two `linePlot` function using a template template parameter to abstract the second argument (`RepresentableFunction` vs `FunctionTree`).

src/utils/Plotter.cpp
```diff
@@ -100,236 +63,336 @@ template <int D> void Plotter<D>::setSuffix(int t, const std::string &s) {
...
     printout(20, std::endl);
 }
 
+/** Parametric plot of a function
+
+    Plots the function func parametrically along the vector A starting
+    from the origin O and returns a vector of function values.
+*/
+template <int D>
+Eigen::VectorXd Plotter<D>::linePlot(const std::array<int, 1> &npts, const RepresentableFunction<D> &func) const {
```
robertodr: Same here, consider using a template template parameter to reduce code bloat. Also, this function doesn't really plot, so maybe consider changing its name?

src/utils/Plotter.cpp
```diff
@@ -100,236 +63,336 @@ template <int D> void Plotter<D>::setSuffix(int t, const std::string &s) {
...
+    Eigen::VectorXd values;
+    if (verifyRange(1)) {
```
robertodr: Shouldn't this throw if a zero range was given?

src/utils/Plotter.cpp
```diff
@@ -100,236 +63,336 @@ template <int D> void Plotter<D>::setSuffix(int t, const std::string &s) {
...
+template <int D>
+void Plotter<D>::linePlot(const std::array<int, 1> &npts, FunctionTree<D> &func, const std::string &fname) {
+    println(20, "----------Line Plot-----------");
     std::stringstream file;
-    file << fname << this->suffix[Plotter<D>::Cube];
-    openPlot(file.str());
-    writeCubeData();
-    closePlot();
+    file << fname << this->suffix[Plotter<D>::Line];
+    if (verifyRange(1)) {
```
robertodr: You could refactor to use the functions returning Eigen vectors inside these plotters.

src/utils/Plotter.cpp
stigrj: Hmm, the difference between the two is actually `const` vs non-`const`. `FunctionTree` is a sub-class of `RepresentableFunction`, but for technical reasons it is not `const` upon function evaluation. I will have a go at `const`-ifying it, as it conceptually should be.

robertodr: Ugh, I totally missed the `const`-ness. I probably should refrain from reviewing at night... Anyway, it's still probably worthwhile using a template template, in case this function ends up being used with parameters that are not in the `RepresentableFunction` hierarchy.

src/utils/Plotter.cpp
stigrj: I removed the Eigen vector returners altogether. Unnecessary and never used.

src/utils/Plotter.cpp
stigrj: I managed to `const`-ify the `FunctionTree::evalf` so the `FunctionTree` instance is no longer necessary.

PR #93, opened 2019-07-19
MRCPP-1.0.0

stigrj:
This PR finalizes version 1.0.0 of MRCPP.

The changes mainly involve printing routines and documentation.

- Separate convenience printing routines into free functions, e.g. `Printer::printHeader() -> print::header()`
- Cleanup printing macros, changing `MSG_FATAL() -> MSG_ABORT()`
- Remove MPI non-blocking routine `mpi::isend_tree`, might be put back in later versions
- Name change `timer.getWallTime() -> timer.elapsed()`
- `Timer` no longer have to be explicitly stopped before calling `elapsed()`


src/utils/details.cpp
```diff
@@ -29,6 +29,38 @@
...
+        line[len] = p[len];
+        len++;
+    }
+    p[len] = 0;
+    return atoi(p);
+}
+
+/** @brief returns the current memory usage of this process, in kB */
+int get_memory_usage() {
+    FILE *file = fopen("/proc/self/statm", "r");
```
robertodr: Does this work on macOS?

stigrj: Good question. @gitpeterwind ?

src/utils/mpi_utils.cpp
```diff
@@ -125,23 +124,22 @@ template <int D> void recv_tree(FunctionTree<D> &tree, int src, int tag, MPI_Com
...
 
     Timer t2;
     sTree.rewritePointers();
-    t2.stop();
-    println(10, " Time rewrite pointers       " << std::setw(30) << t2);
+    println(10, " Time rewrite pointers       " << std::setw(30) << t2.elapsed());
 #endif
 }
 
+/*
```
robertodr: Just remove the commented out portion?

src/utils/details.cpp
gitpeterwind: I haven't a Mac to try. (But I remember that the /proc/self/statm "directory" is special; it is not a regular directory in the sense that it is located on a disk.) . Otherwise, do you have any better measure of total memory usage? It is a very necessary diagnostic when developing/testing, so even if it cannot be used on Mac, I would prefer keep it.

stigrj: I will add a back door returning -1 if the "file" doesn't exist

src/utils/mpi_utils.cpp
stigrj: done

src/utils/details.cpp
robertodr: Yes, it should be kept. I don't know other methods to measure total memory consumption. The alternative of modifying `malloc`/`free` or writing custom allocators is too time consuming to implement.

PR #97, opened 2019-09-05
Norm

gitpeterwind:
Contain a bug fix. The "lastNode" was wrongly defined in rewritepointers. Important only when a tree is first received through MPI, then nodes are added, (which is not done normally).

Also routines used for calculation of norms, and product of norm of two functions.

Everything ready to merge. Will not change behaviour of established things.

### DIFF
--- a/src/treebuilders/multiply.cpp
+++ b/src/treebuilders/multiply.cpp
@@ -297,7 +297,7 @@ template <int D> double dot(FunctionTree<D> &bra, FunctionTree<D> &ket) {
  * @param[in] bra Bra side input function
  * @param[in] ket Ket side input function
  *
- * If exact=true: the grid of bra and ket MUST be equal.
+ * If exact=true: the grid of ket MUST include the grid of bra.
  * If exact=false: does not at any time read the coefficients individually.
  * The product is done for the end nodes of the bra multiplied by the nodes from the
  * ket with either the same idx, or using a lower scale and assuming uniform
@@ -312,13 +312,13 @@ template <int D> double node_norm_dot(FunctionTree<D> &bra, FunctionTree<D> &ket
     double valB[ncoef];
     int nNodes = bra.getNEndNodes();
 
-    if (exact and nNodes != ket.getNEndNodes()) { MSG_ABORT("Trees must have same grid"); }
     for (int n = 0; n < nNodes; n++) {
         FunctionNode<D> &node = bra.getEndFuncNode(n);
         const NodeIndex<D> idx = node.getNodeIndex();
-        FunctionNode<D> *mwNode = static_cast<FunctionNode<D> *>(ket.findNode(idx));
         if (exact) {
             // convert to interpolating coef, take abs, convert back
+            FunctionNode<D> *mwNode = static_cast<FunctionNode<D> *>(ket.findNode(idx));
+            if (mwNode == nullptr) { MSG_ABORT("Trees must have same grid"); }
             node.getAbsCoefs(valA);
             mwNode->getAbsCoefs(valB);
             for (int i = 0; i < ncoef; i++) result += valA[i] * valB[i];
--- a/src/treebuilders/multiply.cpp
+++ b/src/treebuilders/multiply.cpp
@@ -23,6 +23,7 @@
  * <https://mrcpp.readthedocs.io/>
  */
 
+#include <Eigen/Core>
 #include "multiply.h"
 #include "MultiplicationCalculator.h"
 #include "PowerCalculator.h"
@@ -32,6 +33,7 @@
 #include "add.h"
 #include "grid.h"
 #include "trees/FunctionNode.h"
+#include "trees/SerialFunctionTree.h"
 #include "trees/FunctionTree.h"
 #include "trees/HilbertIterator.h"
 #include "utils/Printer.h"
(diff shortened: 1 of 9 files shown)


src/treebuilders/multiply.cpp
```diff
@@ -290,6 +292,46 @@ template <int D> double dot(FunctionTree<D> &bra, FunctionTree<D> &ket) {
...
+    int ncoef = bra.getKp1_d() * bra.getTDim();
+    double valA[ncoef];
+    double valB[ncoef];
+    int nNodes = bra.getNEndNodes();
+
+    if (exact and nNodes != ket.getNEndNodes()) { MSG_ABORT("Trees must have same grid"); }
+    for (int n = 0; n < nNodes; n++) {
+        FunctionNode<D> &node = bra.getEndFuncNode(n);
+        const NodeIndex<D> idx = node.getNodeIndex();
+        FunctionNode<D> *mwNode = (FunctionNode<D> *)ket.findNode(idx);
```
stigrj: please use `static_cast`

src/treebuilders/multiply.cpp
```diff
@@ -290,6 +292,46 @@ template <int D> double dot(FunctionTree<D> &bra, FunctionTree<D> &ket) {
...
+            // convert to interpolating coef, take abs, convert back
+            node.getAbsCoefs(valA);
+            mwNode->getAbsCoefs(valB);
+            for (int i = 0; i < ncoef; i++) result += valA[i] * valB[i];
+        } else {
+            // approximate by product of node norms
+            int rIdx = ket.getRootBox().getBoxIndex(idx);
+            assert(rIdx >= 0);
+            const MWNode<D> &root = ket.getRootBox().getNode(rIdx);
+            result += sqrt(node.getSquareNorm()) * root.getNodeNorm(idx);
```
stigrj: `std::sqrt()`

src/trees/FunctionNode.cpp
```diff
@@ -178,6 +178,21 @@ template <int D> void FunctionNode<D>::getValues(VectorXd &vec) {
...
+/** get coefficients corresponding to absolute value of function
+ *
+ * Leaves the original coefficients unchanged. */
+template <int D> void FunctionNode<D>::getAbsCoefs(double *absCoefs) {
+    double *coefsTmp = this->coefs;
+    for (int i = 0; i < this->n_coefs; i++) { absCoefs[i] = coefsTmp[i]; } // copy
+    this->coefs = absCoefs;                                                // swap coefs
+    this->mwTransform(Reconstruction);
+    this->cvTransform(Forward);
+    for (int i = 0; i < this->n_coefs; i++) { this->coefs[i] = abs(this->coefs[i]); }
```
stigrj: `std::abs`

src/trees/MWNode.cpp
```diff
@@ -903,6 +903,25 @@ template <int D> MWNode<D> *MWNode<D>::retrieveNode(const NodeIndex<D> &idx) {
...
 
+/** Gives the norm (absolute value) of the node at the given NodeIndex.
+ *
+ * Recursive routine to find the node with a given NodeIndex. When an EndNode is
+ * found, do not generate any new node, but rather give the value of the norm
+ * assuming the function is uniformly distributed within the node. */
+template <int D> double MWNode<D>::getNodeNorm(const NodeIndex<D> &idx) const {
+    if (this->getScale() == idx.getScale()) { // we're done
+        assert(getNodeIndex() == idx);
+        return sqrt(this->squareNorm);
```
stigrj: `std::sqrt` here and just below

src/trees/SerialFunctionTree.cpp
```diff
@@ -67,8 +67,9 @@ SerialFunctionTree<D>::SerialFunctionTree(FunctionTree<D> *tree, SharedMemory *m
         this->maxNodesPerChunk = (sizePerChunk / this->sizeNodeCoeff / sizeof(double) / 8) * 8;
     }
 
-    this->lastNode = (ProjectedNode<D> *)this->sNodes; // position of last allocated node
-    this->lastGenNode = this->sGenNodes;               // position of last allocated Gen node
+    this->lastNode =
+        (ProjectedNode<D> *)this->sNodes; // position just after last allocated node, i.e. where to put next node
```
stigrj: `static_cast`. Can you also move the comment to the line above to avoid line break from clang-format?

src/treebuilders/multiply.cpp
gitpeterwind: > Do I understand correctly that even `exact=true` is asymmetric in the bra and ket? It's always the bra that defines the grid, and no nodes are generated in the process?

If `exact=true`, it is required that the grid are already defined equal by the calling routine. `if (exact and nNodes != ket.getNEndNodes()) { MSG_ABORT("Trees must have same grid"); }`

gitpeterwind: > please use `static_cast`

done

src/treebuilders/multiply.cpp
gitpeterwind: > `std::sqrt()`

done (also in MWNode.cpp)

src/treebuilders/multiply.cpp
stigrj: I see. It's not a fool proof check, though, as different trees can have the same number of nodes. You could rather check for `nullptr` in the `findNode()` return value

gitpeterwind: done

PR #102, opened 2019-10-14
Conda recipe

bjorgve:


robertodr:
@bjorgve I was told that it's better to keep recipes in one place rather than together with the code itself. Development of the conda package was moved here: https://github.com/nostress-noworries/nostress/tree/master/conda-recipes/mrcpp We can close this PR and move discussion to the nostress repo.
The packages are deployed here: https://anaconda.org/nostress

PR #105, opened 2019-10-21
MRCPP filter dir can now be specified by export

bjorgve:
Now the `mwfilter` directory can be set by the environment variable `MW_FILTER_DIR`. 
This is useful for our conda packaging of `mrcpp`. 

If this variable is not set, the directory is set to the `default`. 

So the current order of priority is:

1. If environment variable is set, use that 
2. Go for default

I think it could be useful for users of the library should be able to specify this,
if so we should set the order of priority to

1. Code defined path to `MW_FILTER_DIR`
2. If environment variable is set, use that
3. Go for default

### DIFF
--- a/src/core/FilterCache.cpp
+++ b/src/core/FilterCache.cpp
@@ -53,7 +53,13 @@ template <int T> FilterCache<T>::FilterCache() {
         default:
             MSG_ERROR("Invalid filter type: " << T);
     }
-    this->libPath = MWFilter::getDefaultLibrary();
+    auto ep = getenv("MRCPP_FILTER_DIR");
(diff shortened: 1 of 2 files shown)


stigrj:
So does this solve all cases? We need something that works both pre-install
```
cd mrchem/build
make
ctest
```
as well as after the source has been nuked
```
cd mrchem/build
make install
cd ...
rm -rf mrchem
```

stigrj:
Is the environment variable picked up at compile time or run time?

bjorgve:
It's picked up at run time

bjorgve:
It will work after the source has been nuked, apart from the `mwfilters` directory, which can be placed wherever you want. We can then access it at run time through `export MW_FILTER_DIR`

PR #106, opened 2019-10-21
Make build from zip or tar.gz work when Git is around

robertodr:
If Git is around but we're not building from the repo, CMake will panic on the `STRIP` command.  This is a quick fix.


robertodr:
Feel free to squash-merge.

PR #107, opened 2019-10-23
Handle filters

robertodr:
This reworks handling of filters according to what we discussed yesterday at the meeting.

- The `get_mw_filter_dir` function takes the place of the `MW_FILTER_DIR` preprocessor macro. It is also responsible for checking the `MRCPP_FILTER_DIR` env-var.
The function gets the location of the library and "walks" upwards from there to the directory holding the filters.
- Filters are always copied/installed to  `share/MRCPP/mwfilters`, either under `<build-dir>` or `<install-dir>`.


stigrj:
So now the `INSTALL_FILTER` option is gone and the filters are always copied both to <build-dir> and <install-dir>?

robertodr:
Correct, but this is still WIP: I am now testing working with a preinstalled MRCPP.

robertodr:
Not quite correct implementation, it would seem.

robertodr:
I think I'll just translate the filters to source files, rather than do contortions with filesystem operations...

robertodr:
It takes ages to compile this on my laptop. I don't think having these as code is a good solution either... :thinking:

robertodr:
@bjorgve  and I are cooking something else. Closing.

PR #110, opened 2019-11-04
Redo handling of filters

robertodr:
I save both the build and install locations for the filters in the `config.h` file. MWFilter (and other classes that use those files) will loop over the list of locations and use the first one that exists.

This supersedes the disaster in #107


stigrj:
Is this one ready?

robertodr:
This is ready.

PR #113, opened 2019-12-10
Minor fixes

robertodr:
Some small stuff we talked about yesterday. Namely:
- Use the standard `_OPENMP` preprocessor definition rather than `HAVE_OPENMP`.
- Transform macros in `config.h` into `constexpr` functions. This should remove the warnings about macros being redefined.

Similar changes should be done to MRChem.


stigrj:
A general problem (also with the current `master` version) is that `PROGRAM_VERSION` and `GIT_REVISION` are not updated unless you run `cmake`. This means that if you commit pure source changes which doesn't trigger `cmake` and then re-build an existing build directory, then the `GIT_REVISION` will be unchanged.

I've seen this in MRChem when comparing outputs from successive revisions, which made the debugging sessions a bit confusing.

Can you think of a way to trigger `cmake` if the git revision changes?

robertodr:
Radovan and I cooked up something along these lines for the book. I'll try
out on Thursday.

stigrj:
How about this: on `master` the version is *always* undefined as `X.Y.Z`, and there should be *no* version tags in the `master` history, only on `release` branches. Then we show the git hash with a dirty mark
```
------------------------------------------------------------

 MRCPP version         : X.Y.Z
 Git branch            : minor-fixes
 Git commit hash       : 49db92a35b342214a831-dirty
 Git commit author     : Roberto Di Remigio
 Git commit date       : Thu Dec 12 12:32:12 2019 +0100

 Linear algebra        : EIGEN v3.3.7
 Parallelization       : OpenMP

------------------------------------------------------------
```

robertodr:
Works for me, but I would reserve `dirty` for uncommitted changes.

stigrj:
`COMMAND ${GIT_EXECUTABLE} describe --abbrev=20 --long --always --dirty --match "NOT A TAG
"`

PR #114, opened 2019-12-13
Regenerate git info

stigrj:
I made a new `version.h.in` file with the `git` info alongside the old `config.h.in`. The `version.h` file is updated every time we run `make`, but it is only included by `Printer.cpp` so it doesn't trigger a complete rebuild as the old `config.h` file did.

Not sure if the naming is appropriate, or if the configure step of `config.h.in` is correct. I went back to the old and simple configure for this file.

### DIFF
--- a/cmake/custom/main.cmake
+++ b/cmake/custom/main.cmake
@@ -7,9 +7,15 @@ set(CMAKECONFIG_INSTALL_DIR "share/cmake/${PROJECT_NAME}")
 
 file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME})
 
+configure_file(
+  ${PROJECT_SOURCE_DIR}/config.h.in
+  ${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/config.h
+  @ONLY
+  )
+
 add_custom_command(
   OUTPUT
-    ${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/config.h
+    ${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/version.h
   COMMAND
     ${CMAKE_COMMAND} -DINPUT_DIR=${PROJECT_SOURCE_DIR}
                      -DTARGET_DIR=${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}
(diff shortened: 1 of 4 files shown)


cmake/custom/main.cmake
```diff
@@ -35,16 +41,24 @@ add_custom_command(
 add_custom_target(
   binary-info
   ALL
+  COMMAND
+    touch ${PROJECT_SOURCE_DIR}/version.h.in
```
robertodr: Just to make sure we're platform-independent:
```
${CMAKE_BINARY_DIR} -E touch_nocreate ${PROJECT_SOURCE_DIR}/version.h.in
```

cmake/custom/main.cmake
```diff
@@ -35,16 +41,24 @@ add_custom_command(
...
   DEPENDS
-    ${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/config.h
+    ${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/version.h
   )
+
 # See here for the reason why: https://gitlab.kitware.com/cmake/cmake/issues/18399
 set_source_files_properties(${PROJECT_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/config.h
   PROPERTIES
     GENERATED 1
   )
```
robertodr: These lines are not needed anymore.

stigrj: Only for `config.h`? So for `version.h` it should still be there?

robertodr: Yes, it should be there only for `version.h`. `config.h` is generated with `configure_file` and had the `GENERATED` property implicitly.

cmake/custom/main.cmake
stigrj: I get this error
```
stig@pilates :: src/mrcpp/build-git ‹regen-git-hash*› » make                            148 ↵
-- Setting option BUILD_STATIC_LIBS: False
-- Using Eigen3: /usr (version 3.3.7)
-- Found packages:
 * OpenMP

-- Configuring done
-- Generating done
-- Build files have been written to: /home/stig/src/mrcpp/build-git
[  1%] Generating include/MRCPP/version.h
-- Git branch: regen-git-hash
-- Git last commit hash: 8eba1a069a50406cd55f-dirty
-- Git last commit author: Stig Rune Jensen
-- Git last commit date: Thu Dec 12 18:31:42 2019 +0100
make[2]: execvp: ./.: Permission denied
make[2]: *** [CMakeFiles/binary-info.dir/build.make:57: CMakeFiles/binary-info] Error 127
make[1]: *** [CMakeFiles/Makefile2:361: CMakeFiles/binary-info.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

robertodr: If you do `touch` rather than `touch_nocreate`?

stigrj: same

robertodr: OK, I don't understand. Let's keep the original form given that it works.

PR #120, opened 2020-01-11
Some changes needed by VAMPyR

robertodr:
While working on MRChemSoft/vampyr#32 I needed to make some small changes to MRCPP.


stigrj:
Why is the `patch` coverage still checked? :angry:

robertodr:
Now updated :cat2:

PR #122, opened 2020-01-23
add function override for project

Gabrielgerez:


robertodr:
@Gabrielgerez you need to run the code formatter, otherwise the bot gets cranky.

PR #127, opened 2020-02-10
Variable prec

gitpeterwind:
Defines (for nodes)  maxSquareNorm and maxWSquareNorm, which are the max value among one node and its descendants. Those are not set by default, so methods that use them must set them first (using makeMaxSquareNorms()). Using those norms instead of the direct node norms is safer, as the direct norms can be larger at a higher scale, thereby the refinement can miss sharp values.
Wavelet.h is adapted to be able to use them.
apply.cpp alows operator application using a precision weighted by norm given in a separate tree.
multiply.cpp give a new method for multiplying, which does not necessarily use the entire union grid if no significant values are present there.
Will be used by the new faster mrchem Poisson application.
(also defined addition of absolute values for trees; not really needed now, but could be useful anyway)

### DIFF
--- a/src/treebuilders/WaveletAdaptor.h
+++ b/src/treebuilders/WaveletAdaptor.h
@@ -49,47 +49,47 @@ protected:
     bool multiplicationSplit = false;
 
     bool splitNode(const MWNode<D> &node) const override {
-      if(multiplicationSplit and this->precTrees.size()>0){
-	auto multPrec = 1.0;
-        if(this->precTrees.size()!=2)std::cout<<this->precTrees.size()<<" ERROR "<<std::endl;
-        auto &pNode0 = precTrees[0]->getNode(node.getNodeIndex());
-        auto &pNode1 = precTrees[1]->getNode(node.getNodeIndex());
-        double maxW0=std::sqrt(pNode0.getMaxWSquareNorm());
-        double maxW1=std::sqrt(pNode1.getMaxWSquareNorm());
-        double maxS0=std::sqrt(pNode0.getMaxSquareNorm());
-        double maxS1=std::sqrt(pNode1.getMaxSquareNorm());
-        if(pNode0.isGenNode()){
-            maxW0 = 0.0;
-            maxS0 = std::sqrt(std::pow(2.0, D * node.getScale()) *pNode0.getSquareNorm());
-        }
-        if(pNode1.isGenNode()){
-            maxW1 = 0.0;
-            maxS1 = std::sqrt(std::pow(2.0, D * node.getScale()) *pNode1.getSquareNorm());
-        }
-        //The wavelet contribution (in the product of node0 and node1) can be approximated as
-        multPrec =  maxW0*maxS1 +  maxW1*maxS0 + maxW0*maxW1 ;
-
-        // Note: this never refine deeper than one scale more than input tree grids, because when wavelets are zero for both input trees, multPrec=0
-        // In addition, we force not to refine deeper than input tree grids
-	if (multPrec > this->prec and not (pNode0.isLeafNode() and pNode1.isLeafNode())) {
-	  return true;
-	} else {
-	  return false;
-	}
-      }else{
-        auto precNorm = 1.0;
-        if(this->precTrees.size() >0) precNorm = 0.0;//initialize
-        for (int i = 0; i < this->precTrees.size(); i++) {
-            auto &pNode = precTrees[i]->getNode(node.getNodeIndex());
-            auto n = node.getScale();
-            if(not pNode.isGenNode() and pNode.getMaxSquareNorm()>0.0){
-                precNorm = std::max(precNorm, std::sqrt(pNode.getMaxSquareNorm()));
-            }else{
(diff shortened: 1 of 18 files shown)


stigrj:
The `example/` files should probably not have been commited?

src/treebuilders/WaveletAdaptor.h
```diff
@@ -38,13 +38,58 @@ template <int D> class WaveletAdaptor : public TreeAdaptor<D> {
...
     double prec;
     double splitFac;
+    std::vector<FunctionTree<D> *> precTrees;
+    bool multiplicationSplit = false;
 
     bool splitNode(const MWNode<D> &node) const override {
-        return node.splitCheck(this->prec, this->splitFac, this->absPrec);
+        if (multiplicationSplit and this->precTrees.size() > 0) {
+            auto multPrec = 1.0;
+            if (this->precTrees.size() != 2) std::cout << this->precTrees.size() << " ERROR " << std::endl;
```
stigrj: `MSG_ERROR("Invalid precTree size: " << this->precTrees.size());`

gitpeterwind:
> The `example/` files should probably not have been commited?

Why not? they show example of how to use the new features

src/treebuilders/WaveletAdaptor.h
gitpeterwind: Actually we should still discuss the details: I do not understand all the features that are "jumped over". 
What is "Splitfac" used for? 
In principle one could have more than two functions, but when is this used/useful?

stigrj: `splitFac` is explained in the [`crop()`](https://mrcpp.readthedocs.io/en/latest/mrcpp_api/mwfunctions.html#_CPPv4N5mrcpp12FunctionTree4cropEddb) function. I don't think we ever use anything else than `splitFac = 1.0`

stigrj:
Can you do a rebase on `master`? I tried it and there is only one simple conflict to resolve in multiply.cpp. Take the following steps (I could do it, but then I also steal all your commits :slightly_smiling_face: )
```
git fetch upstream
git rebase upstream/master
# resolve conflict in multiply.cpp
git add src/treebuilders/multiply.cpp
git rebase --continue
# this should not give any other conflicts
git push -f origin variable-prec
```

src/trees/MWTree.cpp
```diff
@@ -488,6 +488,18 @@ template <int D> std::ostream &MWTree<D>::print(std::ostream &o) {
     return o;
 }
 
+/** set values for maxSquareNorm in all nodes  */
+template <int D> void MWTree<D>::makeMaxSquareNorms() {
+    NodeBox<D> &rBox = this->getRootBox();
+    MWNode<D> **roots = rBox.getNodes();
+    int DepthMax = 100, slen = 0;
+    ProjectedNode<D> *stack[DepthMax * 8];
```
stigrj, line 496: Are these used?

src/trees/MWNode.cpp
```diff
@@ -58,6 +59,7 @@ MWNode<D>::MWNode()
     setIsLooseNode();
 
     clearNorms();
+    maxSquareNorm = -1.0;
```
stigrj, line 62: This was already initialized to -1.0

src/trees/MWNode.cpp
```diff
@@ -49,6 +49,7 @@ MWNode<D>::MWNode()
         : tree(nullptr)
         , parent(nullptr)
         , squareNorm(-1.0)
+        , maxSquareNorm(-1.0)
```
stigrj, line 52: `maxWSquareNorm` should also be initialized?

src/trees/MWNode.cpp
```diff
@@ -68,6 +70,7 @@ MWNode<D>::MWNode(const MWNode<D> &node)
         : tree(node.tree)
         , parent(nullptr)
         , squareNorm(-1.0)
+        , maxSquareNorm(-1.0)
```
stigrj, line 73: same, add `maxWSquareNorm(-1.0)`

src/treebuilders/WaveletAdaptor.h
```diff
@@ -38,13 +39,58 @@ template <int D> class WaveletAdaptor : public TreeAdaptor<D> {
...
 protected:
     bool absPrec;
     double prec;
     double splitFac;
+    std::vector<FunctionTree<D> *> precTrees;
+    bool multiplicationSplit = false;
 
     bool splitNode(const MWNode<D> &node) const override {
-        return node.splitCheck(this->prec, this->splitFac, this->absPrec);
+        if (multiplicationSplit and this->precTrees.size() > 0) {
```
stigrj, line 53: I think we can make a new `Adaptor` class for this case

stigrj: I will do this separation

src/trees/MWNode.cpp
gitpeterwind: done

src/trees/MWNode.cpp
gitpeterwind: done

stigrj:
I have made a PR to your fork. You probably don't get notifications unless you click the "watch" button in your repo?

stigrj (approved):

PR #129, opened 2020-02-25
Add apply_on_unit_cell method

bjorgve:
Periodic extension for `mrcpp` including negative scale operators and far/near field applies.

We zero center the unit cell [-a, a]. We added FarField and NearField applies to the operators.  

We added negative scale operators, where we move the point of view of the operator to a negative scale relative to
the unit-cell. This is an alternative to increasing the reach of the operator. We do this by adding fake parent nodes
for the input and output functions. Then exploiting the symmetries that comes from a periodic system.

General updates:
- Update max distance for operator reach to handle periodic operators
- Update coordinate and index manipulation to work for zero centered unit-cells
- Add Cartesian product function, these are useful for n-dimensional nested loops
- New and better split node in the OperatorAdaptor

Negative scale operators:

- Made negative scale nodes (parent nodes for the root).
- Update `HilbertIterator` to work for parent nodes.
- Update `mwTransform` to work for the parent nodes
- Use relative depth in `ConvulutionCalculator`

Far field applies:
- Make near field and far field apply
- Add in_unit_cell check 

Link to a mindmap showing the entire pull request:
https://app.mindmup.com/map/_free/2021/08/c0ae4480f8fe11eb876237459bc5a0d8

### DIFF
--- a/src/trees/MWTree.cpp
+++ b/src/trees/MWTree.cpp
@@ -188,15 +188,15 @@ template <int D> void MWTree<D>::incrementNodeCount(int scale) {
 template <int D> void MWTree<D>::decrementNodeCount(int scale) {
     int depth = scale - getRootScale();
     if (depth < 0) {
-        assert(depth < this->nodesAtNegativeDepth.size());
+        assert(-depth - 1 < this->nodesAtNegativeDepth.size());
         this->nodesAtNegativeDepth[-depth - 1]--;
-        assert(this->nodesAtNegativeDepth[depth - 1] >= 0);
-        if (this->nodesAtNegativeDepth[-depth - 1] == 0 and this->nodesAtNegativeDepth.size() > 1) { this->nodesAtNegativeDepth.pop_back(); }
+        assert(this->nodesAtNegativeDepth[-depth - 1] >= 0);
+        if (this->nodesAtNegativeDepth[-depth - 1] == 0 and this->nodesAtNegativeDepth.size() > 1) this->nodesAtNegativeDepth.pop_back();
     } else {
         assert(depth < this->nodesAtDepth.size());
         this->nodesAtDepth[depth]--;
         assert(this->nodesAtDepth[depth] >= 0);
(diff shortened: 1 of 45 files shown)


stigrj:
What's the status on this?

bjorgve:
@stigrj I think it's ready.

src/trees/MWTree.cpp
```diff
@@ -133,8 +133,16 @@ template <int D> void MWTree<D>::mwTransformDown(bool overwrite) {
             int n_nodes = nodeTable[n].size();
 #pragma omp for schedule(guided)
             for (int i = 0; i < n_nodes; i++) {
+                int reverse = 7;
```
stigrj, line 136: This looks specific for 3D?

src/utils/periodic_utils.cpp
```diff
@@ -33,29 +33,68 @@
...
-            if (translation[i] < 0) translation[i] = (translation[i] + 1) % two_n + two_n - 1;
+        l[i] = idx[i] + two_n / 2;
+        if (l[i] >= two_n) return false;
+        if (l[i] < 0) return false;
+    }
+
+    return true;
+}
+
+template <int D> void indx_manipulation(NodeIndex<D> &idx, const std::array<bool, D> &periodic) {
```
stigrj, line 51: Can you make it `index_manipulation`, no use in saving a single character in the name

bjorgve: Fixed

src/trees/MWTree.cpp
bjorgve: think i fixed it, I moved it into the negative scale test, then replaced it with `int reverse = n_nodes - 1`

stigrj:
Added some small fixes.

I think we can let this one in for now. More changes to come when we put `VAMPyR` on top of this, but that will be a separate PR.

stigrj (approved):

PR #130, opened 2020-02-25
Fix Eigen3 cmake as in MRChem

ilfreddy:


robertodr:
@stigrj maybe a new release?

stigrj:
on it

stigrj:
why this weird `before-git-migration` instead of a proper version number?

robertodr:
The `Eigen3Config.cmake` file is broken with all released versions of Eigen when doing the superbuild with `FetchContent`: https://gitlab.com/libeigen/eigen/-/blob/master/cmake/Eigen3Config.cmake.in

PR #132, opened 2020-03-11
Update compiler flags

stigrj:
Same as https://github.com/MRChemSoft/mrchem/pull/268


robertodr:
I'll go ahead and merge. But I agree with @gitpeterwind that testing Intel with `-O3` might squeeze some more performance out of this.

robertodr:
Let's see how this works with MRChemSoft/mrchem#262

gitpeterwind:
>  testing Intel with -O3

I was thinking about `-fp-model fast` and `-fp-speculation=fast`. 
For the record: we had -O3 at some point, but we got problems with alignment. O3 does not check if the data is aligned. It was a long time ago though, and it might have been solved since then.

stigrj:
I just remember that we for some reason moved from `-O3` (or `-Ofast`, not sure which) to `-O2` some time ago, and I didn't want to summon any ghosts of Chistmas past :ghost:

PR #133, opened 2020-03-11
More flags

stigrj:
This seems to fix the error when combining `debug` and `release`. Tested with both `intel/2019b` and `foss/2019b` on Fram, with mrcpp-`release` and mrchem-`debug`.

- moved `-xHost` and `-march=native` to the common flags
- bumped the compiler options to `fast`, as suggested by @gitpeterwind. Final numbers are different in the 14th digit, but we squeezed another percent off the computation time.
- fixed the GNU segfault in the periodic parts of the test.


robertodr:
I'll do a valgrind run on my branch after rebasing for this and MRChemSoft/mrchem#269

PR #135, opened 2020-03-20
Mapping of a function tree through a lambda function

ilfreddy:
Maps a `FunctionTree` through a g(x) as g(f(x)). This functionality is needed for ZORA to obtain
k(x) = 1/(1-V(x)/2c^2), but it is a nice thing to have in general.

- [X] tree builder for the mapping 
- [X] testing of the tree builder
- [x] in-place mapping
- [x] testing of the in-place mapping

### DIFF
--- a/src/treebuilders/MapCalculator.h
+++ b/src/treebuilders/MapCalculator.h
@@ -1,0 +1,58 @@
+/*
+ * MRCPP, a numerical library based on multiresolution analysis and
+ * the multiwavelet basis which provide low-scaling algorithms as well as
+ * rigorous error control in numerical computations.
+ * Copyright (C) 2020 Stig Rune Jensen, Jonas Juselius, Luca Frediani and contributors.
+ *
+ * This file is part of MRCPP.
+ *
+ * MRCPP is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MRCPP is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with MRCPP.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * For information on the complete list of contributors to MRCPP, see:
+ * <https://mrcpp.readthedocs.io/>
+ */
+
+#pragma once
+
+#include "TreeCalculator.h"
+
+namespace mrcpp {
+
+template <int D> class MapCalculator final : public TreeCalculator<D> {
+public:
+    MapCalculator(FMap fm, FunctionTree<D> &inp)
+            : func(&inp)
+            , fmap(std::move(fm)) {}
+
+private:
+    FunctionTree<D> *func;
+    FMap fmap;
+    void calcNode(MWNode<D> &node_o) override {
+        const NodeIndex<D> &idx = node_o.getNodeIndex();
+        int n_coefs = node_o.getNCoefs();
+        double *coefs_o = node_o.getCoefs();
+        // This generates missing nodes
+        MWNode<D> node_i = func->getNode(idx); // Copy node
+        node_i.mwTransform(Reconstruction);
+        node_i.cvTransform(Forward);
+        const double *coefs_i = node_i.getCoefs();
+        for (int j = 0; j < n_coefs; j++) { coefs_o[j] = fmap(coefs_i[j]); }
+        node_o.cvTransform(Backward);
+        node_o.mwTransform(Compression);
+        node_o.setHasCoefs();
+        node_o.calcNorms();
+    }
+};
+
+} // namespace mrcpp
--- a/src/trees/FunctionTree.cpp
+++ b/src/trees/FunctionTree.cpp
@@ -389,6 +389,34 @@ template <int D> void FunctionTree<D>::multiply(double c, FunctionTree<D> &inp) 
     inp.deleteGenerated();
 }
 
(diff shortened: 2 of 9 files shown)


ilfreddy:
Once the tests pass this is ready to go for my part.

src/treebuilders/MapCalculator.h
```diff
@@ -0,0 +1,58 @@
...
+
+#pragma once
+
+#include "TreeCalculator.h"
+
+namespace mrcpp {
+
+template <int D> class MapCalculator final : public TreeCalculator<D> {
+public:
+    MapCalculator(std::function<double(double)> fm, FunctionTree<D> &inp)
```
robertodr: Can you add a typedef or template alias for the function type? 
```c++
template <typename T, typename U>
using FMap_ = std::function<T(U)>;

typedef FMap_<double, double> FMap;
```

src/treebuilders/MapCalculator.h
```diff
@@ -0,0 +1,58 @@
...
+            : func(&inp)
+            , fmap(fm) {}
```
robertodr: It might be slightly more efficient to do:
```c++
fmap(std::move(fm))
```

src/trees/FunctionTree.cpp
```diff
@@ -389,6 +389,34 @@ template <int D> void FunctionTree<D>::multiply(double c, FunctionTree<D> &inp)
...
+ *
+ * @details The input function will be mapped in-place on the current grid
+ * of the function, i.e. no further grid refinement.
+ *
+ */
+template <int D> void FunctionTree<D>::map(std::function<double(double)> fmap) {
+    if (this->getNGenNodes() != 0) MSG_ABORT("GenNodes not cleared");
+    {
+        int nNodes = this->getNEndNodes();
+#pragma omp for schedule(guided)
```
robertodr: What about using a range-based `for` loop?
```c++
for (auto & node: this->endNodeTable) {
      node.mwTransform(Reconstruction);
      node.cvTransform(Forward);
     for (auto & coef : node.getCoefs()) {
            coef = fmap(coef);
      }
      node.cvTransform(Backward);
      node.mwTransform(Compression);
      node.calcNorms();
}
```
I am unsure about the OpenMP `pragma`-s play out with this though, @stigrj opinions?

stigrj: This should be possible with newer versions of OpenMP. I did a quick test and it works on my laptop, but not on Saga and Travis, so let's just keep it as is for now

tests/treebuilders/map.cpp
```diff
@@ -0,0 +1,107 @@
...
+    const double prec = 1.0e-4;
+    double alpha = 1.0;
+    double beta = 50.0;
+    double beta_ref = 100.0;
+
+    double pos_c[3] = {-0.25, 0.35, 1.05};
+    auto pos = details::convert_to_std_array<double, D>(pos_c);
+
+    GaussFunc<D> inp_func(beta, alpha, pos);
+    GaussFunc<D> ref_func(beta_ref, alpha, pos); // LUCA: weird segfault if ths becomes a GaussPoly
```
stigrj: Do we need to address this comment?

src/trees/FunctionTree.cpp
robertodr: Bummer.

src/treebuilders/MapCalculator.h
ilfreddy: And why would that be better? I'm in general not so keen on typedefs. I find they make the code more obscure to read.

src/treebuilders/MapCalculator.h
ilfreddy: OK. But I doubt efficiency plays a role at this level. This is called once per function.

src/trees/FunctionTree.cpp
ilfreddy: I have copied the function from the in-place multiplication, and the `#pragma` stems from that one. I will let @stigrj evaluate if that should stay there or not.

tests/treebuilders/map.cpp
ilfreddy: It is strictly speaking not connected to this PR. I copied the code from the multiplication test and while changing I forgot to change the `ref_func` from `GaussPoly` to `GaussFunc` and got a segfault. I just thought it was wierd. That's why I left the cmment there.

src/treebuilders/MapCalculator.h
ilfreddy: At any rate, where should the typedef be defined?

robertodr: If I want to change the input and output types of the function I change the `typedef` not N  different places. I don't know where you should put the definition.

ilfreddy: @stigrj any recommendations about where to ad the typedef?

stigrj: I guess that could be `api/constants.h` or `api/mrcpp_declarations.h`

ilfreddy: Done. I see there is a lot of `std::function<double(const  coord<D> r)>`
Should it undergo the same typdef-ification? (not in this PR)

stigrj:
A bit too quick there :) Missing `#include <functional>`

ilfreddy:
> A bit too quick there :) Missing `#include <functional>`

Where?

stigrj:
In `mrcpp_declarations.h`

stigrj:
Need final approval from @robertodr. Don't worry about CircleCI, not configured yet

tests/treebuilders/map.cpp
stigrj: I'm not able to reproduce this. I changed this line to
```
GaussPoly<3> ref_func = inp_func * inp_func;
```
but got no errors or segfaults, and it passed through valgrind.

stigrj:
Please avoid merges

ilfreddy:
> Please avoid merges

Ops. Sorry! But What should I do if I need to include a change made by somebodyelse, before I push?
I just issued a `git pull origin map` and there came the merge.

PR #142, opened 2020-08-28
Require python 3.6 to run setup

bjorgve:
I don't know if this is necessary, but I figured it might be a good idea with some consistency.


robertodr:
No need for this check here. Python is used only to generate documentation in MRCPP.

bjorgve:
OK

PR #147, opened 2020-09-22
Export compiler flags with MRCPP CMake target

robertodr:
Puts the compiler flags in the `INTERFACE` of the MRCPP CMake target. The result is that doing `target_link_libraries(<my-target> PUBLIC MRCPP::mrcpp)` will compile `<my-target>` with the same flags as MRCPP.


stigrj (approved):
thx

PR #149, opened 2020-09-26
More fixes for conda package

robertodr:
- Allow to opt out from adding the architecture-dependent `-march=native`/`-xHost` to the `CMAKE_CXX_FLAGS` set. This can be achieved with `--arch-flags=false` in the `setup` script. 
- Do not export compiler flags used to build MRCPP to clients (reverting my latest commit)
- OpenMP and MPI are in the `BUILD_INTERFACE`, _i.e._ only inherited by targets built within the MRCPP project.


stigrj (approved):
Good. Should we add the same flag to MRChem? Is this change compliant with a patch release?

robertodr:
Yes and yes. Please open an mrchem issue and assign it to me.

PR #150, opened 2020-09-30
Encapsulate OMP

stigrj:
The old `_OPENMP` flag has been replaced by `MRCPP_HAS_OMP` in order to distinguish between building MRCPP with/without OMP and building the host program with/without OMP. With this it is possible to compile MRCPP and the host with different OMP support, e.g. serial host with threaded MRCPP, and the threads are correctly picked up only in the library routines. 

Also added option to set `mrcpp::set_max_threads(N)` to limit the number of threads accessible by MRCPP, thus freeing up threads to be used in the host program. Default behavior not changed, i.e. `mrcpp::max_threads = OMP_NUM_THREADS`.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -39,4 +39,6 @@ inline constexpr auto mwfilters_install_dir() noexcept {
 
 #cmakedefine HAVE_MPI
 
+#cmakedefine MRCPP_HAS_OMP
+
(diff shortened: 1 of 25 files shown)


robertodr (approved):
Nice cleanup! Something similar coming up also for MPI?

PR #151, opened 2020-10-01
Encapsulate MPI

stigrj:
Has been tested to work for:

- Serial client -- serial MRCPP (works)
- Serial client -- parallel MRCPP (works as one would expect, client can use MRCPP as serial lib and no MPI parallelization is gained)
- Parallel client -- serial MRCPP (works as long as there's no reference in client to MRCPPs parallel API, the client's parallelization will be completely independent of MRCPP)
- Parallel client -- parallel MRCPP (works, but need to juggle client's `HAVE_MPI` vs `MRCPP_HAS_MPI`)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -5,12 +5,14 @@ add_dependencies(mrcpp mrcpp-info)
 target_compile_definitions(mrcpp
   INTERFACE
     $<INSTALL_INTERFACE:USING_MRCPP>
-  PUBLIC
-    $<$<TARGET_EXISTS:MPI::MPI_CXX>:HAVE_MPI>
   )
 
 if (TARGET OpenMP::OpenMP_CXX)
   target_compile_definitions(mrcpp PUBLIC MRCPP_HAS_OMP)
+endif()
+
+if (TARGET MPI::MPI_CXX)
+  target_compile_definitions(mrcpp PUBLIC MRCPP_HAS_MPI)
 endif()
 
 target_include_directories(mrcpp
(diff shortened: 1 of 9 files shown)


stigrj:
This means that a fully parallel (`--omp --mpi`) MRCPP can now be used in clients with any level of parallelization; serial, omp, mpi or hybrid (as long as it's correctly implemented in the client, ofc)

robertodr (approved):

src/CMakeLists.txt
```diff
@@ -5,14 +5,16 @@ add_dependencies(mrcpp mrcpp-info)
 target_compile_definitions(mrcpp
   INTERFACE
     $<INSTALL_INTERFACE:USING_MRCPP>
-  PUBLIC
-    $<$<TARGET_EXISTS:MPI::MPI_CXX>:HAVE_MPI>
   )
 
 if (TARGET OpenMP::OpenMP_CXX)
   target_compile_definitions(mrcpp PUBLIC MRCPP_HAS_OMP)
```
robertodr, line 11: To let the host know that MRCPP is OpenMP-enabled:
```cmake
set_target_properties(mrcpp
    PROPERTIES
         MRCPP_HAS_OMP TRUE
    )
```

src/CMakeLists.txt
```diff
@@ -5,14 +5,16 @@ add_dependencies(mrcpp mrcpp-info)
...
 endif()
 
+if (TARGET MPI::MPI_CXX)
+  target_compile_definitions(mrcpp PUBLIC MRCPP_HAS_MPI)
```
robertodr, line 15: To let the host know that MRCPP is MPI-enabled:
```cmake
set_target_properties(mrcpp
  PROPERTIES
      MRCPP_HAS_MPI TRUE
  )
```

PR #152, opened 2020-10-02
Some micro optimizations uncovered by MAQAO

stigrj:
Inlining of frequently visited functions


robertodr (approved):
Is it always guaranteed that functions defined in the header are included? Maybe add `inline` too?

robertodr:
How do you run MAQAO?

stigrj:
Pff, it was part of a workshop I did a few weeks ago, and all the tools sort of blend together :slightly_smiling_face: But I think MAQAO was one of the simpler tools, you download a free precompiled binary and attach it to your executable, and it produces a html semi-graphical output, much like ARM's performance reports.

stigrj:
The way I understand it is that functions defined _in_ the class definition are implicitly `inline`. This doesn't necessarily guarantee that the function is _actually_ inlined by the compiler, but I don't think adding the `inline` keyword makes a difference. I might be wrong

robertodr:
Yes, you're right. inline wouldn't do anything special here. And the
compiler decides what to do anyway

PR #162, opened 2020-11-21
Use std::array in NodeIndex

stigrj:


src/trees/NodeIndex.h
```diff
@@ -33,117 +33,78 @@
...
+        return NodeIndex<D>(this->N + 1, l);
+    }
 
-    inline NodeIndex<D> &operator=(const NodeIndex<D> &idx);
-    inline bool operator==(const NodeIndex<D> &idx) const;
-    inline bool operator!=(const NodeIndex<D> &idx) const;
+    // comparisons
+    bool operator!=(const NodeIndex<D> &idx) const { return not(*this == idx); }
+    bool operator==(const NodeIndex<D> &idx) const {
+        bool out = (this->N == idx.N) ? true : false;
```
robertodr, line 60: I think you can skip the ternary if here, `this-> N == idx.N` already gives `true` or `false`.

src/trees/NodeIndex.h
```diff
@@ -33,117 +33,78 @@
...
     int getScale() const { return this->N; }
     int getTranslation(int d) const { return this->L[d]; }
-    int *getTranslation() { return this->L; }
-    const int *getTranslation() const { return this->L; }
 
+    // reference getters
+    int &operator[](int d) { return this->L[d]; }
+    const int &operator[](int d) const { return this->L[d]; }
+    std::array<int, D> &getTranslation() { return this->L; }
+    const std::array<int, D> &getTranslation() const { return this->L; }
```
robertodr, line 77: Why return by reference? There is the setter to set the translation, this should be just:
```c++
std::array<int, D> getTranslation() const { return this->L; }
```
I think.

src/trees/NodeIndex.h
```diff
@@ -33,117 +33,78 @@
...
+
+    // ostream printer
     friend std::ostream &operator<<(std::ostream &o, const NodeIndex<D> &idx) { return idx.print(o); }
```
robertodr, line 80: You can unfriend this one too. Make `print` a `public` method and declare the overload outside the class:
```c++
template <int D>
std::ostream &operator<<(std::ostream &o, const NodeIndex<D> &idx) { return idx.print(o); }
```

robertodr (approved):
Nice refactor! Some minor issues here and there.

robertodr:
Well, Minazo is useless right now. Does this affect the Python bindings in VAMPyR?

stigrj:
Thanks for the feedback. Yes, I guess this will break vampyr...

PR #163, opened 2020-11-23
Update MWTree.cpp

gitpeterwind:
Sizes were overflowed if > 8GB


robertodr (commented):
It would probably be better to have the three functions appearing in this product return `long long int` directly, rather than casting after the fact.

gitpeterwind:
> It would probably be better to have the three functions appearing in this product return `long long int` directly, rather than casting after the fact.

Possibly, but that would involve changing a lot more code: we use almost only int in the code. And the three functions separately are well below the overflow limit.

robertodr (approved):
Fair enough :smile_cat:

PR #165, opened 2021-01-25
Tree coeff

gitpeterwind:
Additional functionalities to be used by the "fast_rotate" of MRChem. Should not change behaviour of other parts.
(also added "noalias" to the Eigen matrix multiplication: it is faster (does not buffer output))

### DIFF
--- a/src/treebuilders/ConvolutionCalculator.cpp
+++ b/src/treebuilders/ConvolutionCalculator.cpp
@@ -339,16 +339,16 @@ template <int D> void ConvolutionCalculator<D>::tensorApplyOperComp(OperatorStat
         if (oData[i] != nullptr) {
             Eigen::Map<MatrixXd> op(oData[i], os.kp1, os.kp1);
             if (i == D - 1) { // Last dir: Add up into g
-                g += f.transpose() * op;
+                g.noalias() += f.transpose() * op;
             } else {
-                g = f.transpose() * op;
+                g.noalias() = f.transpose() * op;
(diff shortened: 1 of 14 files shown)


src/treebuilders/ConvolutionCalculator.cpp
```diff
@@ -339,16 +339,16 @@ template <int D> void ConvolutionCalculator<D>::tensorApplyOperComp(OperatorStat
         if (oData[i] != nullptr) {
             Eigen::Map<MatrixXd> op(oData[i], os.kp1, os.kp1);
             if (i == D - 1) { // Last dir: Add up into g
-                g += f.transpose() * op;
+                g.noalias() += f.transpose() * op;
```
stigrj, line 342: Nice catch. Does this make a difference?

gitpeterwind: I did not test so much, but I think specially the Coulomb potential is significantly faster.

stigrj: Excellent! Have you investigated the effect of the `transpose()`?

stigrj (approved):
Good stuff

src/treebuilders/ConvolutionCalculator.cpp
gitpeterwind: The transpose is fine: it is a "soft" transpose, i.e. it does not actually move any data.

PR #166, opened 2021-01-25
Node allocator

stigrj:
Hopefully this doesn't break anything, but some more testing is required :crossed_fingers: 

- Separate some functionality into `tree_utils` free functions
- Name change: `Serial(Function/Operator)Tree` -> `(Operator/Projected/Gen)NodeAllocator`
- Encapsulate `NodeAllocator` functionality


gitpeterwind (approved):
Good you have the courage to try to clean a bit in the Tree family, I still get confused by all the trees. At least SerialTree and SerialFunctionTree are gone now!
I guess if you have checked that the code does not crash, this should be safe.

robertodr:
I think VAMPyR needs an update after this.

stigrj:
Ok, some more tests have passed, so I think it is safe to merge. Yes, vampyr is lagging behind and currently not working, according to @bjorgve, even the official versions `vampyr-0.1` with `mrcpp-1.3.6`.

PR #167, opened 2021-02-01
Update FunctionTree.cpp

gitpeterwind:
makeTreefromCoeff for prec<0


stigrj (approved):
The `prec < 0` behavior is a bit weird here, as you already pointed out, but that needs to be sorted out later when the mrchem|mrcpp interface is adjusted.

PR #169, opened 2021-02-25
Refactor node

stigrj:
Massive simplification of nodes:

- `ProjectedNode` and `GenNode` removed, now there's only `FunctionNode` and `OperatorNode`
- Only one `NodeAllocator` class which is used for both Functions and Operators
- Initialize nodes into pre-allocated memory using placement `new` operator
- Reverted the ugly `threadSafeGenChildren()`, seems to work correctly now with OMP lock after the bugfix some time ago


gitpeterwind:
Nice! that was a big one.
I do not remember the "bug with the lock initialization" that has been found. Did you test with intel compiler? (wasn't it Intel compiler that caused the problems?)

stigrj:
Hmm, I thought it was foss that caused problems, I will double check. We found an initialization bug with the OMP locks that caused the Conda installations to fail.

gitpeterwind:
not 100% sure it was Intel...

gitpeterwind:
when you have introduced the placed new, `new (child_p) FunctionNode<D>(*this, cIdx);` , do you have added a corresponding delete somewhere?

stigrj:
I get a crash with Intel now. Investigating... Also checking the `delete`

gitpeterwind:
Actually I do not understand the advantage of placed new. I remember we had something like that in the early days, and then removed it.

stigrj:
It calls the constructor, so the node initialization can be moved to a more natural place. Not sure a dedicated `delete` is necessary, but I see now that the destructor should be called explicitly to destroy the OMP lock, which is initialized in the constructor.
 
The Intel crash seems to be triggered by MPI and not OMP :thinking: but it was introduced somewhere in this PR. Could still be the OMP locks, need to dig deeper.

gitpeterwind:
I think there should be an explicit delete for every new. The memory is double allocated there.

stigrj:
How did you check this back then? Valgrind says this is clean

gitpeterwind:
I remember we struggled a lot with making all these destructors consitent. If Valgrind says it is ok, it probably is... but I'd like to understand why :)

stigrj:
Back then we tried to make the `SerialTree` much more than what it finally ended up as, so maybe there was a requirement that made it more complicated? The procedure seems straightforward enough, the `new` on the buffer must be matched by a `delete` on the buffer, while the placement `new` is really just a call to the object's constructor, which should be matched by a manual call to the destructor, no delete: https://stackoverflow.com/questions/8918791/how-to-properly-free-the-memory-allocated-by-placement-new

gitpeterwind:
Yeah, probably I misunderstood the placement new. ... maybe that was why I struggled so much then :)

gitpeterwind:
How does the logic work for a tree without coeff? Should we have a flag in MWTree to tell if the tree has coeff or so? We cannot do the same operations on a tree with coeff and without. (I know this was not done properly before either, and does not necessarily need to be completed in this PR)

gitpeterwind:
I guess a tree is either with or without coeff, there is no reason to have only som coeff? Then we can also make all the flags invisible, since they are implicit from the tree we are on.

stigrj:
Yes, my plan was to attack this issue next :slightly_smiling_face: My plan is to make a separate `GridTree` with `GridNodes` which can be used as a blueprint to generate `FunctionTrees`.  The `GridTree` should also contain some MPI features for domain decomposition, so that it can spit out different grids depending on MPI rank. My first application of this would be in the XC functional evaluation.

gitpeterwind:
About XC functional, I think this can be done easily at the node level: one can store the nodes together with their neighbors in the bank, so that the derivatives are easily available too. To treat one node at a time is not a problem (it is done in rotate).
I could try this, but if you prefer you can follow your strategy.

stigrj:
I will be busy with other things next week, so you might as well do it.

gitpeterwind:
Do you want to merge this now? 
If there are no known problems (what about the the Intel  version?) it is fine to have this simplified version as new base.

stigrj:
No, Intel is not working properly, I will try to figure out why during the weekend.

I also got this error with `intel/2019b` on Betzy with the master versions of MRChem and MRCPP, so I guess we need to do more testing with Intel going forward:
```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Calculation             : Compute initial energy
 Method                  : DFT (PBE0)
 Precision               : 1.00000e-04
 Localization            : On
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Error: ~MWTree(), line 66: Nodes at depth != 1 -> 22396

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 50642 RUNNING AT b2240
=   EXIT CODE: 11
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
   Intel(R) MPI Library troubleshooting guide:
      https://software.intel.com/node/561764
===================================================================================
```

gitpeterwind:
Do you get this error also with omp=1?

stigrj:
It crashed at the same place, but without the `~MWTree()` error message.

stigrj:
It runs correctly on a smaller `alkane-010` system, but fails consistently on `alkane-040`.

gitpeterwind:
Which place does it crash (does it run for some time first?)? Which tree?

stigrj:
Haven't looked too carefully yet, but it happens in "Compute initial energy", so after the orbitals are constructed in the SAD guess. And just to clarify, this is not related to this PR, it fails also on MRChem-master + MRCPP-master.

On the other hand, I think I have a lead on the Intel OMP lock issue which _is_ related to this PR

stigrj:
Ok, this seems to run correctly also with Intel

gitpeterwind:
Well done, those things are always difficult to trace back.

gitpeterwind (approved):

PR #170, opened 2021-03-08
.clang format, longer lines and EndNodes in AppendTree

gitpeterwind:
Small changes: 
.clang_format with ColumnLimit:     200
FunctionTree.appendTreeNoCoeff now also writes EndNodeTable and sets EndNode flags.

### DIFF
--- a/src/trees/FunctionTree.cpp
+++ b/src/trees/FunctionTree.cpp
@@ -23,7 +23,7 @@
  * <https://mrcpp.readthedocs.io/>
  */
 
- #include "FunctionTree.h"
+#include "FunctionTree.h"
 
 #include <fstream>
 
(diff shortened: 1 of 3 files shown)


src/trees/FunctionTree.cpp
```diff
@@ -657,11 +658,26 @@ template <int D> void FunctionTree<D>::appendTreeNoCoeff(MWTree<D> &inTree) {
...
+            branchstack.push_back(thisNode);
+            while (branchstack.size() > 0) {
+                MWNode<D> *branchNode = branchstack.back();
+                branchstack.pop_back();
+                if (branchNode->getNChildren() > 0) {
+                    for (int i = 0; i < branchNode->getNChildren(); i++) {
+                        branchstack.push_back(branchNode->children[i]);
+                    }
+                } else this->endNodeTable.push_back(branchNode);
+            }
```
stigrj, line 680: There is a `resetEndNodeTable()` function in MWTree which you could call after the tree is fully constructed., assuming the endNodeTable does not have to up to date _during_ this process. It will be less efficient, though, as it will traverse the tree one more time.

stigrj (approved):
Thanks!

PR #171, opened 2021-03-24
Update license header

stigrj:


gitpeterwind:
I know I am alone with this opinion, but want to express it anyway:
The first lines of a file are the most visible part, and one expect to see the most important information there. Like explanation about what the code is doing and organized. Instead the license text is only of interest for a lawyer, and there is close to zero chance that anybody will ever bother about it. Legally it is enough to have the license somewhere, and even without license, there are only microscopic chances that it causes any problems. Instead one has to scroll past all this text to get to the interesting parts.
So my vote goes to remove all license info from all the c++ files.

stigrj:
I have no strong opinion on this. Scrolling past this doesn't bother me too much, and it adds a bit of "professional" feel to the project. The legal aspect does not worry me one bit. @ilfreddy @robertodr ?

ilfreddy (approved):

PR #173, opened 2021-09-05
Changes for upcoming VAMPyR update

stigrj:
- There's now only one `TreeIterator` class with a parameter for parser type (Hilbert or Lebesgue) instead of the old derived classes `HilbertIterator` and `LebesgueIterator`. The main reason for this is for clarity when using it as the old class names were not very descriptive.
- Removed the `Gaussian`'s property of being "periodic", with a `GaussExp` member representing it's periodic images. Now there are just regular non-periodic Gaussians and a free function `periodify(Gaussian<D>)` (formerly known as `make_gaussian_periodic`) which returns a semi-periodic version of the Gaussian in the form of a `GaussExp`. So when working in a periodic setting you should first `periodify` your Gaussian and then continue with the resulting `GaussExp`. Not sure if this retains all the functionality of the old version, as I do not remember the ideas behind the old setup.
- Some simple cleanups in the Gaussian classes.


src/functions/function_utils.cpp
```diff
@@ -47,8 +51,8 @@ namespace mrcpp {
  * integral is conserved with respect to the integration limits.
  *
  */
-template <int D> std::shared_ptr<GaussExp<D>> function_utils::make_gaussian_periodic(const Gaussian<D> &inp, const std::array<double, D> &period, double nStdDev) {
-    auto gauss_exp = std::make_shared<GaussExp<D>>();
+template <int D> GaussExp<D> function_utils::periodify(const Gaussian<D> &inp, const std::array<double, D> &period, double nStdDev) {
```
bjorgve, line 54: Add `#include "functions/function_utils.h"` in  `api/Gaussians`. We need this in `mrchem`.

stigrj:
@bjorgve can you think of any issue with the removal of the periodic Gaussians when it comes to its use in MRChem?

bjorgve:
I'm looking into it now.

src/functions/function_utils.cpp
stigrj: Actually, perhaps we should just re-introduce it as a member function of the Gaussian? I like this syntax better:
```
auto periodic_exp = gfunc.periodify(period);
```

bjorgve:
It didn't break any of my periodic tests. So I believe we are safe.

bjorgve (approved):
I'm happy with this

stigrj:
Great! I would still like to move the operator scale stuff from the MRA into the operators themselves, if possible. Looking into it now.

src/functions/function_utils.cpp
bjorgve: I agree

stigrj:
Ok @bjorgve I think I managed to move the `operator_scale` `operator_reach` and `operator_cutoff` stuff out of `MRA` and into the `MWOperator`. Not fully confident that nothing breaks in periodic MRChem, especially with the `apply_(near/far)_field` stuff.

Old way of initializing periodic operator:
```
MultiResolutionAnalysis<3> MRA(pbc_world, basis);

// For negative scale operator
MRA.setOperatorScale(-5);
MRA.setOperatorReach(2);
MRA.setOperatorCutoff(1);

// For reaching operator:
MRA.setOperatorScale(0);
MRA.setOperatorReach(33);
MRA.setOperatorCutoff(32);

PoissonOperator(MRA, prec);
```

New way:
```
MultiResolutionAnalysis<3> MRA(pbc_world, basis);

// Negative scale operator
PoissonOperator(MRA, prec, -5, 1);

// Reaching operator
PoissonOperator(MRA, prec, 0, 32);
```

bjorgve:
This is a nice change, this means we can choose a different reach/scale for the `Poisson` and `Helmholtz` operator. I have to find a clever way to set the reach in `mrchem`. But that's future Magnars problem. I'm happy with this pull request, go ahead and merge if it's ready.

stigrj:
Ok, I think this is good now :crossed_fingers:

PR #174, opened 2021-10-01
Rename scaling factor

bjorgve:
I renamed `sf` -> `scaling_factor`. This makes it easier to grep.

Note on `scaling_factor` in `derivatives`: 

Generally in the code the `scaling_factor`, are placed together with `two_n` factor. For derivatives the `two_n` factor is placed in the construction of the Operators. For the `scaling_factor` it is instead
placed in the apply, since we need access to the `apply_direction`.


stigrj:
Of course, apply_dir is necessary

stigrj (approved):

PR #179, opened 2021-10-10
Add convenience constructor for BoundingBox and MRA

stigrj:
Accepts `[0,L]` or `[-L,L]` arguments for world size, where `L` must be a positive integer.


robertodr:
Yes they are 😅

PR #185, opened 2021-12-10
Add Haar basis MW filters

stigrj:
Allows for function representations with Haar basis `k=0`. ABGV derivative should be supported, but no convolutions or any of the special derivative operators, since they require special input matrices, which I didn't bother to generate


bjorgve (approved):
Good

Issue #187, opened 2021-12-27
mrcpp-tests error


susilehtola:
Running mrcpp-tests in Fedora 35 I get
```
$ mrcpp-tests 
Error: integrate(), line 279: Integration out of bounds
===============================================================================
All tests passed (7885 assertions in 59 test cases)

```

PR #188, opened 2021-12-28
Catch

stigrj:
- Had to update Catch to circumvent compilation error on Fedora 35
- Fixed out-of-bounds error in Polynomial integration, see #187  (non-intrusive error, it just printed an error message but continued correctly)


gitpeterwind (approved):
Looks great :)

PR #190, opened 2021-12-29
Fix array overflow

### LINKED ISSUE #189: Buffer overflow?
The random crash in https://github.com/MRChemSoft/mrchem/issues/389 occurs at https://github.com/MRChemSoft/mrcpp/blob/029e0b72383c01ba17552cf202cdc2a850a71249/src/trees/NodeAllocator.cpp#L344

This line looks suspect to me, since looking at the loop structure
https://github.com/MRChemSoft/mrcpp/blob/029e0b72383c01ba17552cf202cdc2a850a71249/src/trees/NodeAllocator.cpp#L342-L346
the index may have already run over the length of the array.

My quick fix would be
```
diff --git a/src/trees/NodeAllocator.cpp b/src/trees/NodeAllocator.cpp
index 0443213..ec0e99f 100644
--- a/src/trees/NodeAllocator.cpp
+++ b/src/trees/NodeAllocator.cpp
@@ -341,8 +341,8 @@ template <int D> int NodeAllocator<D>::findNextOccupied(int sIdx) const {
     bool endOfStack = (sIdx >= this->topStack);
     while (posIsAvailable and not(endOfStack)) {
         sIdx++;
-        posIsAvailable = (this->stackStatus[sIdx] == 0);
         endOfStack = (sIdx >= this->topStack);
+        posIsAvailable = (!endOfStack) || (this->stackStatus[sIdx] == 0);
     }
     assert(sIdx >= 0);
     assert(sIdx < this->stackStatus.size());
```
but I'm not sure if this is how the function is supposed to work

susilehtola:
Closes #189

### DIFF
--- a/src/trees/NodeAllocator.cpp
+++ b/src/trees/NodeAllocator.cpp
@@ -339,10 +339,12 @@ template <int D> int NodeAllocator<D>::findNextOccupied(int sIdx) const {
     assert(sIdx < this->stackStatus.size());
     bool posIsAvailable = (this->stackStatus[sIdx] == 0);
     bool endOfStack = (sIdx >= this->topStack);
-    while (posIsAvailable and not(endOfStack)) {
-        sIdx++;
-        posIsAvailable = (this->stackStatus[sIdx] == 0);
+    while (1) {
         endOfStack = (sIdx >= this->topStack);
+	if(endOfStack) break;
+        posIsAvailable = (this->stackStatus[sIdx] == 0);
+	if(not posIsAvailable) break;
+        sIdx++;
     }
     assert(sIdx >= 0);
     assert(sIdx < this->stackStatus.size());


susilehtola:
This appears to solve the random crash on Fedora.

src/trees/NodeAllocator.cpp
```diff
@@ -339,10 +339,11 @@ template <int D> int NodeAllocator<D>::findNextOccupied(int sIdx) const {
...
     bool posIsAvailable = (this->stackStatus[sIdx] == 0);
     bool endOfStack = (sIdx >= this->topStack);
-    while (posIsAvailable and not(endOfStack)) {
-        sIdx++;
-        posIsAvailable = (this->stackStatus[sIdx] == 0);
+    while (1) {
         endOfStack = (sIdx >= this->topStack);
+        if(endOfStack) break;
+        posIsAvailable = (this->stackStatus[sIdx] == 0);
+        if(!posIsAvailable) break;
```
stigrj, line 346: I guess `sIdx` still needs to be incremented? Looks like an infinite loop

susilehtola: Yeah sorry, I didn't port the patch properly. Fix coming

susilehtola: Now I copied the real patch that I used in the Fedora package that's already run 46 iterations successfully when it used to crash after 2-6

stigrj (approved):
Excellent, thanks a lot @susilehtola! I wonder why this bug hasn't bit us before :thinking:

susilehtola:
Since this causes a crash, I would recommend releasing 1.4.1 asap :)

PR #192, opened 2022-01-05
Fix node stack buffer overflow, take 2

stigrj:
@susilehtola can you please test if this version runs correctly on your Fedora?


susilehtola:
@stigrj appears to work fine

PR #194, opened 2022-01-14
Fix compile failure on Fedora rawhide arising from missing includes

susilehtola:
`std::unique_ptr` is defined in `<memory>` which needs to be `#include`d to avoid compile failures.


susilehtola:
This is with GCC 12 btw

stigrj (approved):
thx

PR #196, opened 2022-01-23
Cleanups in convolution operators

stigrj:
This will give VAMPyR the ability to build `ConvolutionOperator` from arbitrary `GaussExp<1>`

- Remove unnecessary intermediate class `GreensKernel`
- `ConvolutionOperator` now accepts generic `GaussExp<1>` kernel constructor argument
- `OperatorTreeVector = std::vector<OperatorTree *> -> std::vector<std::unique_ptr<OperatorTree>>`
- `MWOperator.band_max = Eigen::VectorXi -> std::vector<int>`
- Remove `kern_mra` and `oper_mra` from `MWOperator` class, now computed on-the-fly
- Add `MRA<D>` member to `MWOperator`, where `D` is the dimension of the functions we are operating on.
- Moved `getKernelMRA()` and `getOperatorMRA()` from MRA class to `MWOperator`
- Moved `calcMinDistance()` and `calcMaxDistance()` from operator classes to MRA


robertodr:
For `OperatorTreeVector`, does the 2nd CTOR of `unique_ptr` suffice? I recall wrapping smart pointers into STL containers as having some pitfalls...

stigrj:
Which is the 2nd CTOR of `unique_ptr`?

robertodr:
https://en.cppreference.com/w/cpp/memory/unique_ptr

```c++
template <
    class T,
    class Deleter
> class unique_ptr<T[], Deleter>;
```

stigrj:
Sorry, I do not understand the problem

robertodr:
> You mean it's unsafe to do the following?
> 
> ```
> std::vector<std::unique_ptr<OperatorTree>> oper_vec;
> auto o_tree = std::make_unique<OperatorTree>(mra, prec);
> oper_vec.push_back(std::move(o_tree));
> ```

yes, or rather, there used to be a warning against using smart pointers like that. But some Googling shows it's pre-C++11 thing.

src/operators/DerivativeConvolution.cpp
```diff
@@ -0,0 +1,70 @@
...
+namespace mrcpp {
+
+/** @returns New DerivativeConvolution object
+ *  @param[in] mra: Which MRA the operator is defined
+ *  @param[in] pr: Build precision, closeness to delta function
+ *  @details This will project a kernel of a single differentiated
+ *  gaussian with exponent sqrt(10/build_prec).
+ */
+template <int D>
+DerivativeConvolution<D>::DerivativeConvolution(const MultiResolutionAnalysis<D> &mra, double prec)
```
robertodr, line 39: Why derivative convolution? Is this computing a convolution integral after applying the derivative?

src/operators/IdentityConvolution.cpp
```diff
@@ -0,0 +1,70 @@
...
+namespace mrcpp {
+
+/** @returns New IdentityConvolution object
+ *  @param[in] mra: Which MRA the operator is defined
+ *  @param[in] pr: Build precision, closeness to delta function
+ *  @details This will project a kernel of a single gaussian with
+ *  exponent sqrt(10/build_prec).
+ */
+template <int D>
+IdentityConvolution<D>::IdentityConvolution(const MultiResolutionAnalysis<D> &mra, double prec)
```
robertodr, line 39: Same question here about the "convolution" in the name.

src/operators/DerivativeConvolution.cpp
stigrj: It's the derivative operator written as a convolution with a differentiated delta function as kernel, approximated by a narrow p-type Gaussian. So "Convolution" indicates that it's derived from the `ConvolutionOperator` class, as opposed to the "real" derivative operators (`ABGVOperator`, `PHOperator` and `BSOperator`) which are not convolutions but `DerivativeOperators`. The `DerivativeConvolution` is not really useful, but it was an interesting take on our derivative problems of the past.

src/operators/IdentityConvolution.cpp
stigrj: Same as above, it's the identity operator written as a convolution with a delta function kernel, approximated by a narrow Gaussian

robertodr: OK, thanks for the explanation. Can you add this explanation to the Doxygen documentation string? 😸

src/operators/DerivativeConvolution.cpp
robertodr: OK, thanks for the explanation. Can you add this explanation to the Doxygen documentation string? 😸

stigrj: It's there in the header file:
https://mrcpp.readthedocs.io/en/latest/mrcpp_api/mwoperators.html#_CPPv4I_iEN5mrcpp21DerivativeConvolutionE

robertodr: Yes, but the comment sounds clearer to me than what is there right now and it also explains the difference with the other operators, including that this one is not really useful in comparison.

bjorgve (approved):
It looks good to me

stigrj:
I will update the doc string before merging

PR #197, opened 2022-04-22
Negative scale on single root node

stigrj:
- [x] Switch to single root PBC
- [x] Make tests pass
- [x] Set restrictions on world parameters when running PBC
- [x] Cleanup

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -6,6 +6,7 @@ list(APPEND _examples
     multiplication
     poisson
     scf
+    pbc
     mpi_shared_tree
     mpi_send_tree
     mpi_matrix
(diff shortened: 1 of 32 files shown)


stigrj:
I can't remember, @bjorgve , were we supposed to keep the `[-1,1]` possibility for the zero-scale reaching operator? Or should PBC _always_ be done on `[0,1]`?

bjorgve:
Both `[-1, 1]` and `[0, 1]` is better, but `[0, 1]` is fine.

bjorgve (approved):
It looks good to me, I think there should be a sanity check in the Poisson and Helmholtz operator constructors. The reach/root constructors should only work for periodic `mra`. 

I suggest we wait with the merge until I'm finished with periodic mrchem.

bjorgve:
@stigrj can you think about a solution for `VAMPyR`s build prec in the operator constructions?

bjorgve:
The `apply_near_field` does not appear to work as expected for negative scale operators.

stigrj:
Dead end

PR #198, opened 2022-09-12
first working version with parallel in mrcpp

gitpeterwind:
The long term goal is that all the mpi treatment is done "transparently" in mrcpp. The user must simply use "OrbitalVector" (aka MPI_FuncVector), and all the operations on that vector will be done in parallel. Mrcpp should not need to know much about orbitals, only that they are complex functions.

The parallel.cpp, Bank.cpp moved from mrchem to mrcpp
QMFunctions and Orbitals are merged and moved from mrchem to mrcpp into CplxFunc

For now only additional files in mrcpp, so should not affect other parts.

rank is now the indice in the vector (was mpi rank of owner)
renamed orb_rank and orb_size to wrk_rank and wrk_size
... and plenty of small changes.

The mrchem version that can use this mrcpp is here: https://github.com/gitpeterwind/mrchem/tree/vecmult

Renamed the old CplxFunc class to ComplexFunction


stigrj (commented):
Can you remove the file `ComplexFunction.h`, it is currently not used. If the `rankID` no longer refers to rank ID but position in vector, it should be renamed to `index` or `id` or something.

stigrj (approved):

PR #201, opened 2023-01-18
Reintroduce build_prec for vampyr

bjorgve:
Not sure if this is how we want to add it. But we need it in vampyr.


src/operators/ConvolutionOperator.cpp
```diff
@@ -51,7 +51,7 @@ template <int D>
 ConvolutionOperator<D>::ConvolutionOperator(const MultiResolutionAnalysis<D> &mra, GaussExp<1> &kernel, double prec)
         : MWOperator<D>(mra, mra.getRootScale(), -10) {
     int oldlevel = Printer::setPrintLevel(0);
-
+    this->build_prec = prec;
```
stigrj, line 54: This should be put in the initializer list above

stigrj (approved):

PR #205, opened 2023-02-21
bug fix

gitpeterwind:
In a rotation of orbitals, when the transformation matrix U is zero, the output was just a copy of the input, instead of zero.
U is zero in the calculation of the Helmholtz argument when canonical orbitals are used (?).


stigrj:
Yes, with canonical orbitals the rotation matrix becomes zero in the iterations where we actually diagonalize, since the diagonal is subtracted from the Fock matrix in the rotation `F_ij - Lambda_ij`. In iterations where we don't explicitly diagonalize, the rotation matrix is slightly perturbed from zero.

stigrj (approved):

PR #207, opened 2023-05-08
Add getChildPts and getQuadPts methods to MWNode for VAMPyR

bjorgve:
- Implement `getChildPts()` and `getQuadPts()` methods in MWNode class.
- These methods return child and quadrature points as a vector of vectors of doubles.
- Both methods call the corresponding `getPrimitive*Pts()` methods to obtain the points in `Eigen::MatrixXd` format and then convert the points to `std::vector<std::vector<double>>`.

Also bugfix on col/row mixup in getPrimitiveQuadPts

### DIFF
--- a/src/trees/MWNode.cpp
+++ b/src/trees/MWNode.cpp
@@ -672,6 +672,21 @@ template <int D> void MWNode<D>::getPrimitiveQuadPts(MatrixXd &pts) const {
     for (int d = 0; d < D; d++) pts.row(d) = sFac * (roots.array() + static_cast<double>(l[d]));
 }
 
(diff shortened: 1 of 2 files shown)


bjorgve:
When I think about it, we can probably do this wrapping from the `vampyr` side. If so it should only contain the bugfix.

robertodr:
pybind11 understands how to wrap Eigen objects to Numpy objects though, so copying into a vector-of-vectors shouldn't be needed

bjorgve:
I'll do this on the `vampyr` side. I'll do a PR with the bugfix only.

PR #209, opened 2023-06-14
Changing occupancies from integer to double

lratcliff:
Changing orbital occupancies from integer to double, so that we can do fractional occupancies e.g. for core hole calculations


ilfreddy:
Superseded by #243

PR #212, opened 2023-07-24
Vecmult

gitpeterwind:
Extend range of tree objects that can be multiplied by a representable function


robertodr (approved):
Looks good to me, for the little I can understand of how MPI is used in MRCPP

PR #213, opened 2023-08-09
Catchv3 support in tests.

### LINKED ISSUE #211: Port mrcpp to Catch v3
mrcpp appears to be still using Catch v2, whereas most Linux distributions have already migrated to Catch v3.

The upstream migration guide is at
https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md

susilehtola:
Closes #211 

This code follows the recent Catch v3 port in https://github.com/wavefunction91/IntegratorXX/


cmake/DownloadProject.cmake
robertodr, line 1: Is this file necessary? `FetchContent` is part of CMake nowadays, indeed we do use it already extensively in MRCPP/MRChem.

tests/CMakeLists.txt
robertodr, line 1: Can you rewrite the bit of the CMake script where you fetch to something like this? `MakeAvailable` will populate and add the subdirectory.

https://github.com/MRChemSoft/mrcpp/blob/df9b05402e8d2c88ecda066ef728f3088d57d3e5/external/upstream/fetch_eigen3.cmake

robertodr (changes requested):
Thanks! I'd been wanting to do this for a long time :) Please see my comments on what should be changed.

tests/CMakeLists.txt
susilehtola: Done

cmake/DownloadProject.cmake
susilehtola: Removed

robertodr (approved):
Thank you!

PR #224, opened 2023-10-04
Time evolution operator in 1d

edinvay:
The free particle time Schrodinger operator is ready for use.


examples/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ list(APPEND _examples
     mpi_shared_tree
     mpi_send_tree
     mpi_matrix
+	schrodinger_semigroup1d
```
ilfreddy, line 12: Indentation

examples/schrodinger_semigroup1d.cpp
```diff
@@ -0,0 +1,144 @@
...
+    auto world = mrcpp::BoundingBox<1>(min_scale);
+    auto MRA = mrcpp::MultiResolutionAnalysis<1>(world, basis, max_depth);
+
+    mrcpp::print::footer(0, timer, 2);
+
+    // Time moment:
+    double t = 0.001;
+    int finest_scale = 8;
+
+    mrcpp::TimeEvolutionOperator<1> ReExp(MRA, prec, t, finest_scale, false);
```
ilfreddy, line 38: We name classes with the first letter capitalized, but objects are not capitalized, and possibly with longer names
Example:
`class MyClass`
`MyClass myObject`
So in your case it should be `TimeEvolutionOperator realExpansion`

I see many other places where this convention is for the time being not followed. It is not critical but it is important that the coding style is uniform throughout the mrcpp code. This not only has aesthetic reasons, but it makes the code more readable and quicker to understand if the conventions are followed.

src/core/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ target_sources(mrcpp
     ${CMAKE_CURRENT_SOURCE_DIR}/ObjectCache.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/QuadratureCache.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/ScalingBasis.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/SchrodingerEvolution_CrossCorrelation.cpp
```
ilfreddy, line 13: Same stuff for file names: `SchrodingerEvolutionCrossCorrelation`

src/core/SchrodingerEvolution_CrossCorrelation.cpp
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * For information on the complete list of contributors to MRCPP, see:
+ * <https://mrcpp.readthedocs.io/>
+ */
+
+/*
+ *
+ *
+ *  \date Jul 18, 2009
+ *  \author Jonas Juselius <jonas.juselius@uit.no> \n
```
ilfreddy, line 30: This is definitel not written by Jonas Juselius. Please remove :-)

src/operators/TimeEvolutionOperator.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "utils/math_utils.h"
+
+#include "treebuilders/TimeEvolution_CrossCorrelationCalculator.h"
+
+#include <vector>
+
+
+namespace mrcpp {
+
+template <int D>
```
ilfreddy, line 58: Documentation should be added here

tools/Plotting_data.ipynb
```diff
@@ -0,0 +1,323 @@
+{
```
ilfreddy, line 1: I would rather not have python notebooks in mrcpp

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
```diff
@@ -0,0 +1,475 @@
+{
```
ilfreddy, line 1: Python notebooks should not be in mrcpp

ilfreddy (changes requested):
Nice work!
I've added a few comments. Mostly is style-related issues. 

On top of that yon need to add unit testing to make sure that the functionality you have been adding to the code is not going to be broken by other changes in mrcpp, and a proper documentation.

But I'm looking forward to being able to see this at work!

examples/CMakeLists.txt
edinvay: This depends on the editor one uses to browse the file. In Visual Studio the indentation is fine, and I do not know how to make it the same in all editors.

src/core/SchrodingerEvolution_CrossCorrelation.cpp
edinvay: Done.

examples/schrodinger_semigroup1d.cpp
ilfreddy: Can you rewrite those as being `realExponent` instead of `ReExp`? Then they are both more readable and in line with the style used in mrchem/mrcpp. There might be exceptions, yes but the coding style of both mrchem and mrcpp is as I described. But we can take this up tomorrow at the meeting, in connection with code documentation.

src/operators/TimeEvolutionOperator.cpp
edinvay: Done

tools/Plotting_data.ipynb
edinvay: This one will be removed.

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
edinvay: This is questionable...

examples/CMakeLists.txt
ilfreddy: Well it should not. I guess the settings og your editor should be twaked :-)

src/core/CMakeLists.txt
edinvay: I used to read "The practice of programming" when I was young. It is not a good idea to have long names combined from words starting with capital letters. In this case `_` separates prefix, which makes it better for readability in my humble opinion.

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
ilfreddy: Why? Please explain. 
1. It is definitely not required for mrcpp to work
2. It could at least be made into a python script (not notebook)
Python notebooks clutter the commit history with many unnecessary commits (notebooks metadata change every time the script is run).

examples/CMakeLists.txt
robertodr: Seems you're using 2 tabs instead of one.

src/functions/JpowerIntegrals.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+std::vector<std::complex<double>> & JpowerIntegrals::operator[](int index)
+{
+    if( index < 0 ) index += integrals.size();
+    return integrals[index];
+}
+
+std::vector<std::complex<double>> JpowerIntegrals::calculate_J_power_inetgarls(int l, double a, int M, double treshold)
+{
+    const double pi = 3.1415926535897932384626433832795;
```
robertodr, line 73: You should use the `M_PI` constant defined in the `cmath` header.

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
robertodr: Agree with Luca here that notebooks should not be in MRCPP. I will add to Luca's list that they're truly horrible to code review, despite the integration with ReviewNB. Please make it into a script.

examples/CMakeLists.txt
edinvay: I used firstly one tab in a normal text editor, which turned out to be different from the indention used. However, when I opened it in vs code, it was fine. So my guess is that the others are done in vs code by using space key. Yes, it seems that Stig used spaces before, and I never use space key for indentation.

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
edinvay: I do not think there is any point to run this script more often than once per several years. You used maybe ones and then keep only as a documentation.

src/functions/JpowerIntegrals.cpp
edinvay: Done

src/core/CMakeLists.txt
ilfreddy: There is no right or wrong when it comes to conventions. There is a style that is adopted in a given code. And the style adopded in mrchem/mrcpp is `MyClass myObject` so it should be used throughout the code. If everyone starts adopting a different style the code becomes a mess :-)

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
ilfreddy: The best option, again is a simple python script that does exactly the same stuff. The python notebooks should be kept separate fromm the main code because what both Roberto and me said. Under git they make a mess and they don't belong there.

src/core/CMakeLists.txt
edinvay: What about examples/poisson.cpp line 45?
`mrcpp::PoissonOperator P(MRA, prec);`
Shouldn't it be `mrcpp::PoissonOperator Poisson(MRA, prec);` or `mrcpp::PoissonOperator operatorPoisson(MRA, prec);`.
IPersonallyDislikeThisConventionSoMuchBecauseICannotReadItMyselfNotJustBecauseIThinkItIsJustBetter :)

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
edinvay: The notion `separate` should be properly defined. On the last meeting I was told to use tools folder, and now I am being told differently. On my personal laptop is not a solution. Can you easily find generators of filters and cross correlation coefficients? It has to be on git somewhere.

src/core/CMakeLists.txt
ilfreddy: I'm sure there are places where the convention is not respected, and that should ideally be changed. That's not a good reason avoid using it. I think this is better discussed tomorrow at th meeting so we are all on the same page.

tools/Schrodinger_evolution_cross_correlation_coefficients.ipynb
robertodr: No, you're being told that the notebooks do not belong in MRCPP. It's fine to create a `tools` folder, but please, convert the code in the notebook to a Python script. It's not about how easy it is _for you_ to run these things _now_, it's about how easy you can make it for _other people_  to re-run things in the _next years_.

src/functions/JpowerIntegrals.h
```diff
@@ -0,0 +1,151 @@
...
+    /// @brief creates an array of power integrals
+    /// @param a : parameter a
+    /// @param scaling : scaling level
+    /// @param M : maximum amount of integrals for each l
+    /// @param treshold : lower limit for neglecting the integrals
+    /// @details The array is orginised as a vector ordered as \f$l = 0, 1, 2, \ldots, 2^n - 1, 1 - 2^n, 2 - 2^n, \ldots, -2, -1 \f$.
+    JpowerIntegrals(double a, int scaling, int M, double treshold = 1.0e-15);
+    JpowerIntegrals(const JpowerIntegrals& other);
+
+    ~JpowerIntegrals();
```
robertodr, line 142: You're using data structure that do [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) so I'm pretty sure having an explicit destructor is useless (if not even harmful) here.

src/functions/JpowerIntegrals.h
```diff
@@ -0,0 +1,151 @@
...
+
+    int scaling;  //it is probably not used
+    std::vector<std::vector<std::complex<double>>> integrals;
+
+    std::vector<std::complex<double>> & operator[](int index);
+    std::vector<std::complex<double>> calculate_J_power_inetgarls(int l, double a, int M, double treshold);
```
robertodr, line 148:
```suggestion
    std::vector<std::complex<double>> calculate_J_power_integrals(int l, double a, int M, double threshold);
```

src/functions/JpowerIntegrals.h
```diff
@@ -0,0 +1,151 @@
...
+    /// @param M : maximum amount of integrals for each l
+    /// @param treshold : lower limit for neglecting the integrals
+    /// @details The array is orginised as a vector ordered as \f$l = 0, 1, 2, \ldots, 2^n - 1, 1 - 2^n, 2 - 2^n, \ldots, -2, -1 \f$.
+    JpowerIntegrals(double a, int scaling, int M, double treshold = 1.0e-15);
+    JpowerIntegrals(const JpowerIntegrals& other);
+
+    ~JpowerIntegrals();
+
+    int scaling;  //it is probably not used
+    std::vector<std::vector<std::complex<double>>> integrals;
```
robertodr, line 145: I suggest you use a `Eigen::MatrixXcd` (https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html) instead of a doubly-nested `std::vector` here:
* Eigen allocates the matrix data in one chunk, and I believe you know the number of rows and columns when you start computing the integrals you need.
* The API of the Eigen matrices is more intuitive for index-based operations.
* Eigen can be faster than using `std::vector` directly.

src/functions/JpowerIntegrals.h
```diff
@@ -0,0 +1,151 @@
...
+{
+public:
+    /// @brief creates an array of power integrals
+    /// @param a : parameter a
+    /// @param scaling : scaling level
+    /// @param M : maximum amount of integrals for each l
+    /// @param treshold : lower limit for neglecting the integrals
+    /// @details The array is orginised as a vector ordered as \f$l = 0, 1, 2, \ldots, 2^n - 1, 1 - 2^n, 2 - 2^n, \ldots, -2, -1 \f$.
+    JpowerIntegrals(double a, int scaling, int M, double treshold = 1.0e-15);
+    JpowerIntegrals(const JpowerIntegrals& other);
```
robertodr, line 140: Do you really need an explicit copy-constructor? Can you point to where it's used?

robertodr: I think this class is a "simple" data container and should follow the so-called [rule of zero](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero)

src/functions/JpowerIntegrals.h
robertodr: I see this function is only used in the constructor. It should be behind a `private:` section, so things are kept properly encapsulated.

src/functions/JpowerIntegrals.cpp
```diff
@@ -0,0 +1,98 @@
...
+    return integrals[index];
+}
+
+std::vector<std::complex<double>> JpowerIntegrals::calculate_J_power_inetgarls(int l, double a, int M, double treshold)
+{
+    //const double pi = 3.1415926535897932384626433832795;
+    const std::complex<double> I(0.0, 1.0);  // Imaginary unit
+    const std::complex<double> J_0 = 0.25 * std::exp(-0.25 * I * M_PI) / std::sqrt(M_PI * a) * std::exp(0.25 * I * static_cast<double>(l * l) / a);
+    const std::complex<double> beta(0, 0.5 / a);
+    const std::complex<double> alpha = static_cast<double>(l) * beta;
```
robertodr, line 80: Few comments for these lines:
* These are local variables, so using `const` doesn't do much and, personally, I find it an unnecessary distraction.
* You can use `auto` as return type, if you want to save yourself some typing.
* You can use a literal to get the imaginary unit: https://en.cppreference.com/w/cpp/numeric/complex/operator%22%22i
```c++
using namespace std::complex_literals;

auto J_0 = 0.25 * std::exp(-0.25i * M_PI) / std::sqrt(M_PI * a) * std::exp(0.25i * static_cast<double>(l * l) / a);
```

src/functions/JpowerIntegrals.h
edinvay: It looks that there is no proper git command `git rename treshold threshold`... how can I correct these typos in multiple files?

examples/CMakeLists.txt
edinvay: Done

src/functions/JpowerIntegrals.h
edinvay: At the beginning I wanted to create/allocate two objects of the time evolution operator for real and imaginary part in a separate method, but I had some memory issues, so I had to give up that idea. Now indeed I do not need it anymore.

src/functions/JpowerIntegrals.h
edinvay: Done

src/core/SchrodingerEvolution_CrossCorrelation.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+
+void SchrodingerEvolution_CrossCorrelation::setCCCPath(const std::string &lib) {
+    switch (this->type) {
+        case (Interpol):
+            MSG_ERROR("Not implemented yet filter type: " << this->type);
+            break;
+        case (Legendre):
+            this->path = lib + "/Schrodinger_evolution_cross_correlation_coefficients_Legendre_scaling_type.bin";
```
ilfreddy, line 80: This should be done the same way as the usual cross correlation coeffiients. A binary file will most likely not be portable.

src/functions/JpowerIntegrals.h
```diff
@@ -0,0 +1,152 @@
...
+    //JpowerIntegrals(const JpowerIntegrals& other);
+
+    //~JpowerIntegrals();
+
+    int scaling;  //it is probably not used
+    std::vector<std::vector<std::complex<double>>> integrals;
+
+    std::vector<std::complex<double>> & operator[](int index);
+private:
+    std::vector<std::complex<double>> calculate_J_power_integrals(int l, double a, int M, double threshold);
```
ilfreddy, line 149: Please use the naming convention for class methods. In this case `calculateJpowerIntegrals`

src/functions/special_functions.cpp
ilfreddy, line 1: It would be nice if you could follow the naming conventions here as well
`special_functions` --> `specialFunctions` and so on...

src/functions/special_functions.cpp
```diff
@@ -0,0 +1,98 @@
...
+ * \f]
+ * and \f$ g_{a,b} (x) = 0 \f$ elsewhere.
+ * 
+ * @returns The non-negative value
+ * \f( g_{a,b} (x) \f)
+ * at the specified space coordinate \f$ x \in \mathbb R \f$.
+ * 
+ * 
+ */
+double smooth_compact_function(double x, double a, double b) {
```
ilfreddy, line 90: `smoothCompactFunction`

src/operators/TimeEvolutionOperator.cpp
```diff
@@ -0,0 +1,173 @@
...
+ */
+template <int D>
+TimeEvolutionOperator<D>::TimeEvolutionOperator
+(const MultiResolutionAnalysis<D> &mra, double prec, double time, int finest_scale, bool imaginary, int max_Jpower)
+    : ConvolutionOperator<D>(mra, mra.getRootScale(), -10)   //One can use ConvolutionOperator instead as well
+{
+    int oldlevel = Printer::setPrintLevel(0);
+    this->setBuildPrec(prec);
+    
+    SchrodingerEvolution_CrossCorrelation cross_correlation(30, mra.getOrder(), mra.getScalingBasis().getScalingType() );
```
ilfreddy, line 81: `crossCorrelation`

share/mwfilters/Schrodinger_evolution_cross_correlation_coefficients_Legendre_scaling_type.bin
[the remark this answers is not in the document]
edinvay: Discussed on the meeting and agreed that it is in line with the usual mrcpp practice.

src/functions/special_functions.cpp
edinvay: See math_utils.h, function_utils.h, ....

examples/schrodinger_semigroup1d.cpp
edinvay: How do you like this name:
class MPI_FuncVector
?

src/operators/TimeEvolutionOperator.cpp
edinvay: That's an object and the name is given according to the mrcpp convention.
Check this:
private:
    FunctionData func_data;
    mrcpp::SharedMemory *shared_mem_re;
    mrcpp::SharedMemory *shared_mem_im;
    mrcpp::FunctionTree<3> *re; ///< Real part of function
    mrcpp::FunctionTree<3> *im; ///< Imaginary part of function

    void flushFuncData() {
        this->func_data.real_size = 0;
        this->func_data.imag_size = 0;
        if (this->re != nullptr) {
            this->func_data.real_size = this->re->getNChunksUsed();
            flushMRAData(this->re->getMRA());
        }
        if (this->im != nullptr) {
            this->func_data.imag_size = this->im->getNChunksUsed();
            flushMRAData(this->im->getMRA());
        }
    }

    void flushMRAData(const mrcpp::MultiResolutionAnalysis<3> &mra) {
        const auto &box = mra.getWorldBox();
        this->func_data.type = mra.getScalingBasis().getScalingType();
        this->func_data.order = mra.getOrder();
        this->func_data.depth = mra.getMaxDepth();
        this->func_data.scale = box.getScale();
        this->func_data.boxes[0] = box.size(0);
        this->func_data.boxes[1] = box.size(1);
        this->func_data.boxes[2] = box.size(2);
        this->func_data.corner[0] = box.getCornerIndex().getTranslation(0);
        this->func_data.corner[1] = box.getCornerIndex().getTranslation(1);
        this->func_data.corner[2] = box.getCornerIndex().getTranslation(2);
    }

src/functions/JpowerIntegrals.h
edinvay: This is indeed, one of the very few examples, where I do not follow the usual mrcpp convention. At leat somewhere in the code I would like to use those notations that I can read myself. It makes me upset, when I cannot read my own code. It also slows me down a lot. This piece `calculate_J_power_integrals` I can read and grasp in less than a second. This piece `calculateJpowerIntegrals` I need to process in my head for 10-20 seconds at least.

We already discussed it, but I remind again, sorry. It is not even possible to guess the meaning of this piece `int kp1_d = node.getKp1_d();`, though it is probably in line with the convention, or is it really in line?

As I said, mostly I follow the convention, and if somewhere I don't, it means that I want to be able to read the code I am writing. I am not a good programmer, and so I am not able to grasp things like this  `calculateJpowerIntegrals` or like this `getKp1_d`.

ilfreddy:
Aside naming at this point you need testing before this can be merged. This is mostly for your own sake. You don't want to end up in a situation where some modification in MRCPP will end up breaking the TD part.
As for naming I give up for now, but if everybody keeps their own "conventions" the code base becomes a mess. So it will have to adhere to the standard at some point.

ilfreddy (approved):

PR #225, opened 2023-11-02
Const-correctness in cplxfunc::deep_copy

robertodr:


robertodr:
This was not as quick of a fix as I thought it would be 🤦🏻

ilfreddy:
@robertodr Do you want to proceed with this or shall we close this PR?

robertodr:
It would require way larger changes than I had originally anticipated. It seems basically every file in MRCPP needs to be changed! Something to keep in mind when working on the complex algebra changes though.

PR #231, opened 2023-11-30
Node xc

gitpeterwind:
Tools for the nodewise DFT , mostly to make (zero bandwidth) derivative of a node.

Also orthogonalize MPI_FuncVector aka OrbitalVector.

Automatic determination of the right number of omp threads, based on hardware accessibilty. OMP_NUM_THREADS not used for MPI.

Unused memory reserved for GenNodes (GenNodeChunks) are cleaned every time nodes are deallocated: it appeared that the unused memory for GenNodes could be very large, and was kept until the tree was deleted, which sometimes happens only when the entire OrbitalVector was cleared. 

It appeared that the operative system of Betzy had a "memory leak": (some deleted memory was not reused). The solution is to allocate memory in chunks of 32 MB (16MB ws not enough). **NB: orbitals stored on disk with previous versions, can therefore not directly be read by this version.**

### DIFF
--- a/tests/trees/tree_io.cpp
+++ b/tests/trees/tree_io.cpp
@@ -37,7 +37,7 @@ namespace tree_io {
 template <int D> void testProjectFunction();
 
 SCENARIO("FunctionTree IO", "[tree_io], [trees]") {
-    const double prec = 1.0e-4;
+    const double prec = 1.0e-6;
 
     GaussFunc<3> *func = nullptr;
     initialize(&func);


ilfreddy (approved):

src/functions/RepresentableFunction.h
```diff
@@ -74,4 +76,15 @@ template <int D> class RepresentableFunction {
...
+/*
+ * Same as RepresentableFunction, but output a matrix of values
+ * for all points in a node, given its NodeIndex.
+ *
+ */
+class RepresentableFunction_M {
+public:
+    RepresentableFunction_M() {}
+    virtual Eigen::MatrixXd evalf(mrcpp::NodeIndex<3> nIdx) const = 0;
+};
```
stigrj, line 88: Is this class actually used? I see there's a `FunctionTree::integrateEndNodes` function taking such an argument, but I can't find an actual implementation of any `RepresentableFunction_M`

gitpeterwind: Not yet: My original plan was to make a representableFunction_M for representing the XCpotential, so that the entire potential never needed to be stored. However I did not use it at the end. I still think it may/will be useful: in general it is much more efficient to treat all points of a node together, rather than one by one.

ilfreddy (approved):

PR #243, opened 2024-04-24
Changing occupancies from integer to double

msnik1999:
Changing orbital occupancies from integer to double, so that we can do fractional occupancies e.g. for core hole calculations. @lratcliff


ilfreddy:
@lratcliff @msnik1999 Do we still need this PR? In case, can you rebase it and check if it passes the test after the rebase?
There were some issue with the `macos-latest`build and now we instead require `macos-12`, which should be there once you rebase.

stigrj (approved):
Ok, seems to work also within MRChem

PR #246, opened 2024-08-06
Local html documentation with graphs made by Doxygen

ilfreddy:
I was not able to make *sphinx* produce "clickable" diagrams for class inheritance. This PR provides a local workaround: when the documentation is built by *doxygen* it produces both the `xml` stuff required for further processing by *sphinx* and the `html` output. To avoid issues with the `html` produced by *sphinx* later, the native *doxygen* `html` is in the folder `docs/html_local`. It can be inspected by pointing your browser at `docs/html_local/index.html`

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Set up environment
-      uses: conda-incubator/setup-miniconda@v2
+      uses: conda-incubator/setup-miniconda@v3
       with:
           auto-update-conda: true
           auto-activate-base: false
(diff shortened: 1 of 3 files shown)


docs/index.rst
robertodr, line 1: Is this change relevant?

robertodr (approved):

docs/index.rst
ilfreddy: Don't remember right now. I believe it was to obtain the diagrams that stig made. But I can check

ilfreddy: Nope. I guess it was left after an attempt to obtain the doxygen class hierarchy tree, which failed.

robertodr (approved):

PR #249, opened 2024-12-30
Save or read orbitals in text format, portable and readable by MADNESS group

gitpeterwind:
Orbitals can be saved and read in text format (ascii) files. Each end node is saved with info about depth and translation.
Use keyword write_orbitals_txt = T to save in this format.

Also introduced keywords  `bank_per_node` and `use_omp_num_threads` after feedback from Moritz and Roberto. (see documentation)

NB: this PR builds on top of the Four Components PR and is to be used with the corresponding MRChem PR.


src/functions/RepresentableFunction.h
```diff
@@ -38,20 +38,21 @@
 #include "MRCPP/constants.h"
 #include "MRCPP/mrcpp_declarations.h"
 #include "trees/NodeIndex.h"
+#include "MRCPP/utils/math_utils.h"
 
 namespace mrcpp {
 
-template <int D> class RepresentableFunction {
+template <int D, typename T> class RepresentableFunction {
```
robertodr, line 45: Why not have a default for `T` here?
```suggestion
template <int D, typename T = double> class RepresentableFunction {
```

src/functions/RepresentableFunction.cpp
```diff
@@ -110,8 +110,11 @@ template <int D> bool RepresentableFunction<D>::outOfBounds(const Coord<D> &r) c
     return false;
 }
 
-template class RepresentableFunction<1>;
-template class RepresentableFunction<2>;
-template class RepresentableFunction<3>;
+template class RepresentableFunction<1, double>;
```
robertodr, line 113: here we might want to instantiate also for `float` and `std::complex<float>`. Maybe we should have a macro to do it for us, because we'd then be at 12 lines of boilerplate for almost every class. I might have something in my archives...

src/operators/OperatorState.h
```diff
@@ -42,9 +42,9 @@ namespace mrcpp {
 
 #define GET_OP_IDX(FT, GT, ID) (2 * ((GT >> ID) & 1) + ((FT >> ID) & 1))
 
-template <int D> class OperatorState final {
+template <int D, typename T> class OperatorState final {
```
robertodr, line 45: also here we can have a default for T, as the class is marked as `final` and we can't inherit from it anyway
```suggestion
template <int D, typename T = double> class OperatorState final {
```

src/operators/OperatorStatistics.h
```diff
@@ -32,14 +32,14 @@
...
-template <int D> class OperatorStatistics final {
+  template <int D, typename T> class OperatorStatistics final {
 public:
     OperatorStatistics();
     ~OperatorStatistics();
 
     void flushNodeCounters();
-    void incrementFNodeCounters(const MWNode<D> &fNode, int ft, int gt);
-    void incrementGNodeCounters(const MWNode<D> &gNode);
+    void incrementFNodeCounters(const MWNode<D, T> &fNode, int ft, int gt);
```
robertodr, line 41: Actually, I don't think this class should be templated on both `D` and `T` at all. It might be enough to simply template the methods that accept a `MWNode` object :thinking:

src/treebuilders/AdditionCalculator.h
```diff
@@ -30,26 +30,48 @@
...
+        : sum_vec(inp),
+          conj(conjugate) {}
 
 private:
-    FunctionTreeVector<D> sum_vec;
+    FunctionTreeVector<D, T> sum_vec;
+    bool conj;
 
-    void calcNode(MWNode<D> &node_o) override {
+    void calcNode(MWNode<D, double> &node_o) {
```
robertodr, line 43: I believe we can keep the `override` marker here. We don't need to have two implementations for the different scalar types: one can use `if constexpr` to do compile-time, type-level checks nowadays. Of course it's going to be a bit of an ugly function...

src/treebuilders/CrossCorrelationCalculator.cpp
```diff
@@ -77,7 +77,7 @@ template <int T> void CrossCorrelationCalculator::applyCcc(MWNode<2> &node, Cros
         const MWNode<1> &node_a = this->kernel->getNode(idx_a);
         const MWNode<1> &node_b = this->kernel->getNode(idx_b);
 
-        VectorXd vec_a;
+        Eigen::Matrix<double, Eigen::Dynamic, 1> vec_a;
         VectorXd vec_b;
```
robertodr, line 81:
```suggestion
        VectorXd vec_a, vec_b;
```

src/treebuilders/DerivativeCalculator.cpp
```diff
@@ -261,18 +261,19 @@ template <int D> void DerivativeCalculator<D>::applyOperator(OperatorState<D> &o
 
 /** Perform the required linear algebra operations in order to apply an
 operator component to a f-node in a n-dimensional tensor space. */
-template <int D> void DerivativeCalculator<D>::tensorApplyOperComp(OperatorState<D> &os) {
-    double **aux = os.getAuxData();
+template <int D, typename T> void DerivativeCalculator<D, T>::tensorApplyOperComp(OperatorState<D, T> &os) {
+    T **aux = os.getAuxData();
     double **oData = os.getOperData();
+    /*
```
robertodr, line 267: Should the commented-out code be removed altogether? Eigen does have the possibility of offloading to BLAS/LAPACK https://eigen.tuxfamily.org/dox/TopicUsingBlasLapack.html

src/treebuilders/MultiplicationCalculator.h
```diff
@@ -30,23 +30,25 @@
...
+            : prod_vec(inp),
+              conj(conjugate) {}
 
 private:
-    FunctionTreeVector<D> prod_vec;
+    FunctionTreeVector<D, T> prod_vec;
+    bool conj;
 
-    void calcNode(MWNode<D> &node_o) override {
+    void calcNode(MWNode<D, double> &node_o) {
```
robertodr, line 43: Same comment as for the other calculator.

src/treebuilders/SquareCalculator.h
```diff
@@ -29,20 +29,22 @@
...
+            : func(&inp),
+              conj(conjugate) {}
 
 private:
-    FunctionTree<D> *func;
+    FunctionTree<D, T> *func;
+    bool conj;
 
-    void calcNode(MWNode<D> &node_o) override {
+    void calcNode(MWNode<D, double> &node_o) {
```
robertodr, line 42: Same comment here as well

src/treebuilders/add.cpp
```diff
@@ -126,67 +128,143 @@ template <int D> void add(double prec, FunctionTree<D> &out, FunctionTreeVector<
...
+template <int D, typename T> void add(double prec, FunctionTree<D, T> &out, std::vector<FunctionTree<D, T> *> &inp, int maxIter, bool absPrec, bool conjugate) {
+    FunctionTreeVector<D, T> inp_vec;
     for (auto &t : inp) inp_vec.push_back({1.0, t});
-    add(prec, out, inp_vec, maxIter, absPrec);
+    add(prec, out, inp_vec, maxIter, absPrec, conjugate);
 }
 
-template void add<1>(double prec,
-                     FunctionTree<1> &out,
+template void add<1, double>(double prec,
```
robertodr, line 137: yes, a macro would go a very long way to reduce the boilerplate also for these instantiations!

src/treebuilders/apply.cpp
```diff
@@ -91,6 +91,64 @@ template <int D> void apply(double prec, FunctionTree<D> &out, ConvolutionOperat
...
+template <int D> void apply(double prec, CompFunction<D> &out, ConvolutionOperator<D> &oper, const CompFunction<D> &inp, ComplexDouble metric[4][4], int maxIter, bool absPrec) {
+
+    ComplexDouble defaultMetric[4][4];
+    for (int i=0; i<4; i++){
+        for (int j=0; j<4; j++){
+            if (i==j) defaultMetric[i][j] = 1.0;
+            else defaultMetric[i][j] = 0.0;
+        }
+    }
+    if (metric == nullptr) {
```
robertodr, line 128: C++17 introduced [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional) to deal with arguments that may or may not be passed. We should use that here.

src/treebuilders/apply.cpp
```diff
@@ -205,6 +263,36 @@ template <int D> void apply(double prec, FunctionTree<D> &out, ConvolutionOperat
...
 
+template <int D, typename T> void apply(double prec, CompFunction<D> &out, ConvolutionOperator<D> &oper, CompFunction<D> &inp, FunctionTreeVector<D, T> *precTrees, ComplexDouble metric[4][4], int maxIter, bool absPrec) {
+    ComplexDouble defaultMetric[4][4];
+    for (int i=0; i<4; i++){
+        for (int j=0; j<4; j++){
+            if (i==j) defaultMetric[i][j] = 1.0;
+            else defaultMetric[i][j] = 0.0;
+        }
+    }
+    if (metric == nullptr) {
```
robertodr, line 274: Same comment about `std::optional` applies also here.

src/treebuilders/apply.cpp
```diff
@@ -227,10 +315,42 @@ template <int D> void apply(double prec, FunctionTree<D> &out, ConvolutionOperat
...
 
+template <int D> void apply_far_field(double prec, CompFunction<D> &out, ConvolutionOperator<D> &oper, CompFunction<D> &inp, ComplexDouble metric[4][4], int maxIter, bool absPrec) {
+    ComplexDouble defaultMetric[4][4];
+    for (int i=0; i<4; i++){
+        for (int j=0; j<4; j++){
+            if (i==j) defaultMetric[i][j] = 1.0;
+            else defaultMetric[i][j] = 0.0;
+        }
+    }
+    if (metric == nullptr) {
```
robertodr, line 330: `std::optional` also here.

src/treebuilders/apply.cpp
```diff
@@ -308,6 +461,46 @@ template <int D> void apply(FunctionTree<D> &out, DerivativeOperator<D> &oper, F
     print::separator(10, ' ');
 }
 
+template <int D> void apply(CompFunction<D> &out, DerivativeOperator<D> &oper, CompFunction<D> &inp, int dir, ComplexDouble metric[4][4]) {
+    //TODO: sums and not only each components independently
```
robertodr, line 465: TODO still relevant?

src/treebuilders/apply.cpp
```diff
@@ -308,6 +461,46 @@ template <int D> void apply(FunctionTree<D> &out, DerivativeOperator<D> &oper, F
...
+        metric = defaultMetric;
+    }
+    for (int icomp = 0; icomp < inp.Ncomp(); icomp++){
+        for (int ocomp = 0; ocomp < 4; ocomp++){
+            if (std::norm(metric[icomp][ocomp]) > MachinePrec) {
+                if (inp.isreal() and (std::imag(metric[icomp][ocomp]) < MachinePrec or inp.Ncomp() == 1) ) {
+                    apply(*out.CompD[ocomp], oper, *inp.CompD[icomp], dir);
+                    if (std::norm(metric[icomp][ocomp] - 1.0) > MachinePrec) {
+                        if(std::imag(metric[icomp][ocomp]) < MachinePrec) out.CompD[ocomp]->rescale(std::real(metric[icomp][ocomp]));
+                        else out.func_ptr->data.c1[ocomp] *= metric[icomp][ocomp]; //TODO: multiply c1 in rescale?
```
robertodr, line 483: TODO still relevant?

src/treebuilders/map.cpp
```diff
@@ -66,12 +66,12 @@ namespace mrcpp {
  *
  */
 template <int D>
-void map(double prec, FunctionTree<D> &out, FunctionTree<D> &inp, FMap fmap, int maxIter, bool absPrec) {
+void map(double prec, FunctionTree<D, double> &out, FunctionTree<D, double> &inp, FMap<double, double> fmap, int maxIter, bool absPrec) {
```
robertodr, line 69: is `map` not relevant when one has `float` or `std::complex` scalar types?

src/treebuilders/multiply.h
```diff
@@ -28,57 +28,65 @@
...
-                            FunctionTree<D> &ket);
+template <int D, typename T> T dot(FunctionTree<D, T> &bra,
+                            FunctionTree<D, T> &ket);
 
-template <int D> double node_norm_dot(FunctionTree<D> &bra,
-                                      FunctionTree<D> &ket,
+template <int D> ComplexDouble dot(FunctionTree<D, ComplexDouble> &bra,
+                            FunctionTree<D, double> &ket);
+template <int D> ComplexDouble dot(FunctionTree<D, double> &bra,
+                            FunctionTree<D, ComplexDouble> &ket);
```
robertodr, line 47: I think it's enough to have one function, but with three template parameters (similarly could be done for all arithmetic functions, actually)
```suggestion
template <int D, typename T, typename U = T, typename V = decltype(std::declval<T>() * std::declval<U>())>
V dot(FunctionTree<D, T> &bra, FunctionTree<D, U> &ket);
```
you can see why I dislike implementing templates in `.cpp` files :sweat_smile: (cough cough @ilfreddy)

Note that in the C++ standard mixing `float` and `std::complex<double>` (or `double` and `std::complex<float>`) is _not_ possible, so the above should be combined with a compile-time exclusion of such a possibility (through SFINAE or something) Another thing I have in my archives.

src/trees/FunctionNode.h
```diff
@@ -32,55 +32,66 @@
...
-    double integrateInterpolating() const;
-    double integrateValues() const;
+    T integrateLegendre() const;
+    T integrateInterpolating() const;
+    T integrateValues() const;
 };
-
-template <int D> double dot_scaling(const FunctionNode<D> &bra, const FunctionNode<D> &ket);
-template <int D> double dot_wavelet(const FunctionNode<D> &bra, const FunctionNode<D> &ket);
+template <int D>
```
robertodr, line 82: again, I think these signatures can be declared only once using the "trick" above.

src/trees/FunctionTree.cpp
```diff
@@ -101,21 +103,274 @@ template <int D> void FunctionTree<D>::allocRootNodes() {
...
+template <int D, typename T> FunctionTree<D, T>::~FunctionTree() {
+    if (this->getNNodes()>0) this->deleteRootNodes();
+}
+
+
+/** @brief Read a previously stusing MADNESS conventions for n, l and index order.ored tree assuming text/ASCII format,
+ *   in a representation
+ * @param[in] file: File name
+ * @note This tree must have the exact same MRA the one that was saved(?)
+ */
```
robertodr, line 115: The docstring here seems a bit mangled. Can you fix it?

src/trees/FunctionTree.cpp
```diff
@@ -101,21 +103,274 @@ template <int D> void FunctionTree<D>::allocRootNodes() {
...
+    int rscale = this->getRootScale(); //root scale of target MRA (MRChem) . NB: negative
+    for (int i = rscale; i < 0; i++) p *= 2;
+    int L = p; //NB for now we assume the world as a cube going from -L to +L and L is a power of 2
+    // We require that the world box size is identical and a power of 2
+    double TXT_thres = 1.0e-14; // threshold for differences in scaling factors
+    for (int d = 0; d < D; d++) {
+        if (std::abs(coord[d][0] + L) > TXT_thres) std::cout<<coord[d][0]<<" "<<L<<std::endl;;
+        if (std::abs(coord[d][0] + L) > TXT_thres) NOT_IMPLEMENTED_ABORT;
+        if (std::abs(coord[d][1] - L) > TXT_thres) std::cout<<coord[d][1]<<" "<<L<<std::endl;;
+        if (std::abs(coord[d][1] - L) > TXT_thres) NOT_IMPLEMENTED_ABORT;
```
robertodr, line 134: I don't understand why the 2nd and 4th statements

src/trees/FunctionTree.cpp
```diff
@@ -586,31 +875,31 @@ template <int D> int FunctionTree<D>::crop(double prec, double splitFac, bool ab
...
-void FunctionTree<D>::makeCoeffVector(std::vector<double *> &coefs,
+template <int D, typename T>
+void FunctionTree<D, T>::makeCoeffVector(std::vector<T *> &coefs,
                                       std::vector<int> &indices,
                                       std::vector<int> &parent_indices,
                                       std::vector<double> &scalefac,
                                       int &max_index,
-                                      MWTree<D> &refTree,
-                                      std::vector<MWNode<D> *> *refNodes) {
+                                      MWTree<D, double> &refTree,
```
robertodr, line 884: here (and some other places) you explicitly fix the scalar type of some function inputs. Why is that?

src/trees/FunctionTree.cpp
```diff
@@ -785,8 +1117,286 @@ template <> int FunctionTree<3>::saveNodesAndRmCoeff() {
...
+            for (int i = 0; i < inp_node.getNCoefs(); i++) { out_coefs[i] = std::imag(inp_coefs[i]); }
+            out_node.calcNorms();
+        }
+    }
+    out->mwTransform(BottomUp);
+    out->calcSquareNorm();
+    return out;
+}
+
+    /*
```
robertodr, line 1196: I think it's fine to delete commented-out code.

src/trees/MWNode.cpp
```diff
@@ -624,28 +645,28 @@ template <int D> double MWNode<D>::getWaveletNorm() const {
...
     int size = this->getKp1_d();
     int start = i * size;
 
     double sq_norm = 0.0;
-#ifdef HAVE_BLAS
-    sq_norm = cblas_ddot(size, &c[start], 1, &c[start], 1);
-#else
-    for (int i = start; i < start + size; i++) { sq_norm += c[i] * c[i]; }
-#endif
+//#ifdef HAVE_BLAS
```
robertodr, line 658: commented-out code can be removed

src/trees/MultiResolutionAnalysis.cpp
```diff
@@ -146,7 +146,12 @@ template <int D> bool MultiResolutionAnalysis<D>::operator==(const MultiResoluti
  *  For more information about the meaning of equality for BoundingBox and ScalingBasis objets, see their respective classes.
  */
 template <int D> bool MultiResolutionAnalysis<D>::operator!=(const MultiResolutionAnalysis<D> &mra) const {
-    return !(*this == mra);
+    if (this->basis != mra.basis) std::cout<<"diff basis "<<this->basis<<std::endl <<"and  "<< mra.basis<<std::endl;
```
robertodr, line 149: are the print statements needed here? if yes, maybe the should use the printer. Also, no need to check the condition twice, just do a non-inline if-statement.

robertodr (changes requested):
* Can you run `clang-format` on all the C++ files this PR changes?
* When giving a default for the scalar type (_e.g._ `typename T = double`) I think it should be enough to do it on the parent classes. This is however just to save some typing :)
* I think we should have compile-time checks on whether the scalar type `T` makes sense or not, _e.g._ only one of `float`, `double`, `std::complex<float>`, `std::complex<double>`. I have some code to do this that we can reuse.
* For the `metric` arguments of some functions in `apply` we may use `std::optional<Eigen::Matrix<std::complex<double>, 4, 4>>`.
* There's a lot more boilerplate in the explicit instantiations. The obvious solution is to _not_ put the implementation in `.cpp` files, but leave everything in the header. @ilfreddy (and maybe @stigrj?) feels strongly against it, so we'll have to make do with some preprocessor macros and/or some metaprogramming tricks.
* I think we should allow to mix scalar types, but it can be left for a future refinement, as it involves some tricky gymnastics with template parameters.

src/utils/math_utils.h
```diff
@@ -67,6 +67,7 @@ double matrix_norm_1(const Eigen::MatrixXd &M);
 double matrix_norm_2(const Eigen::MatrixXd &M);
 
 void apply_filter(double *out, double *in, const Eigen::MatrixXd &filter, int kp1, int kp1_dm1, double fac);
+void apply_filter(ComplexDouble *out, ComplexDouble *in, const Eigen::MatrixXd &filter, int kp1, int kp1_dm1, double fac);
```
robertodr, line 70: I think the two can be unified with `if constexpr`

robertodr (commented):
* in `parallel.h`/`parallel.cpp` I think the implementations for `double` and `std::complex<double>` could be unified, so we reduce a bit the boilerplate (and possibly extend to `float`)

src/functions/RepresentableFunction.h
gitpeterwind: it is already set as default in mrcpp_declarations.h

src/operators/OperatorState.h
gitpeterwind: it is already set as default in mrcpp_declarations.h

src/treebuilders/apply.cpp
gitpeterwind: I rephrased: // To consider: multiply c1 in rescale?
 The "c1 constant" is somewhat new and it is not yet completely integrated into the logic. It is a "soft" multiplicative constant; soft meaning the same as the conjugate flag for a function: not explicitly applied before it is required. 
 So far it is used for the complex spin operators: we do not want to convert all real functions to complex just because there is a complex multiplicative factor in the operator.

src/treebuilders/map.cpp
gitpeterwind: I don't know, but since it is not used I prefer to not implement before we know what we want to use it for.

src/treebuilders/multiply.h
gitpeterwind: This is nice!

src/trees/FunctionNode.h
gitpeterwind: I did not use the trick for the dot_scaling and dot_wavelet, because with the conjugate flags it became messy.

gitpeterwind:
I have now included all the changes you proposed, or left a comment.
It is a pity that I did not use the `if constexpr` from the start. Actually I did, but then went back because it was not compliant with the std14. 
About using float I am sceptical: for small system there is no point, and for large "HPC" systems the error would be too large. Roughly a HPC calculation would take 10¹² operation at least, giving sqrt(10⁻¹²)  = 10⁻⁶ relative numerical error, 6 digits error of about 7 significant digits is not useful.
About reducing the boilerplate instantiations, I do not think it is important: it is not those parts that take time to implement, and they do not reduce code readability (while macros might!).

gitpeterwind:
yes

PR #250, opened 2025-01-28
Remove macos-12 test (temporarily)

Gabrielgerez:
We still need mac build tests


ilfreddy (approved):
We need to get going with the latest modifications, hence we suspend testing on MAC for now. We should resume as soon as possible.

ilfreddy:
Temporary removal of macos-12 testing

PR #255, opened 2025-03-10
Prebuild

gitpeterwind:
Compared to older versions, there was a difference when projecting a representable function (not a function) that created differences in MRChem results in some cases.
"Prebuild" adds a build when a representable is projected.

### DIFF
--- a/src/trees/NodeAllocator.cpp
+++ b/src/trees/NodeAllocator.cpp
@@ -130,7 +130,7 @@ template <int D, typename T> int NodeAllocator<D, T>::alloc(int nNodes, bool coe
 
     // we require that the index for first child is a multiple of 2**D
     // so that we can find the sibling rank using rank=sIdx%(2**D)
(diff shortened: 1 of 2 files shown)


src/trees/NodeAllocator.cpp
```diff
@@ -130,7 +130,7 @@ template <int D, typename T> int NodeAllocator<D, T>::alloc(int nNodes, bool coe
 
     // we require that the index for first child is a multiple of 2**D
     // so that we can find the sibling rank using rank=sIdx%(2**D)
-    if (sIdx % nNodes != 0) MSG_ERROR(" node allocate error");
+    if (sIdx % nNodes != 0) std::cout << "Warning: recommended number of siblings is 2**D" << std::endl;
```
robertodr, line 133: I guess we should use the `MSG_` macros that Stig made :)
```suggestion
    if (sIdx % nNodes != 0) MSG_WARN("Warning: recommended number of siblings is 2**D");
```

PR #257, opened 2025-03-12
change in the definition of dot multiplication for complex "CompFunction"

gitpeterwind:
There are several ways to define the application of a purely imaginary Operator (like momentum) onto a real function, without defining a fully complex function as result.


robertodr (approved):

gitpeterwind:
Thanks Roberto!

PR #267, opened 2025-11-14
Fftw3 - SmoothDerivative

Christian48596:


ilfreddy:
Not active PR. Closing for now. It can be reactivated if need be.

PR #279, opened 2026-04-10
Refactor copy to complex

ilfreddy:
Small refactoring of the `CopyTreeToComplex` and `CopyTreeToReal` methods. 

- Generalized for D=1,2,3
- Changed the signature

To be done:
switch from raw pointers to `unique_pointer`s but this requires a refactoring of `CompFunction`

### DIFF
--- a/src/trees/FunctionTree.cpp
+++ b/src/trees/FunctionTree.cpp
@@ -1206,9 +1206,8 @@ template <int D, typename T> FunctionTree<D, double> *FunctionTree<D, T>::Imag()
  */
 template <int D, typename T>
 template <typename U, typename>
-void FunctionTree<D, T>::CopyTreeToComplex(FunctionTree<D, ComplexDouble> *&outTree) {
-    delete outTree;
-    outTree = new FunctionTree<3, ComplexDouble>(this->getMRA());
+FunctionTree<D, ComplexDouble>* FunctionTree<D, T>::CopyTreeToComplex() {
+    auto* outTree = new FunctionTree<D, ComplexDouble>(this->getMRA());
     std::vector<MWNode<3, double> *> instack;         // node from this
     std::vector<MWNode<3, ComplexDouble> *> outstack; // node from outTree
     outTree->clearEndNodeTable();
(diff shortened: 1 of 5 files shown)


src/trees/FunctionTree.cpp
```diff
@@ -1206,9 +1206,8 @@ template <int D, typename T> FunctionTree<D, double> *FunctionTree<D, T>::Imag()
  */
 template <int D, typename T>
 template <typename U, typename>
-void FunctionTree<D, T>::CopyTreeToComplex(FunctionTree<D, ComplexDouble> *&outTree) {
-    delete outTree;
-    outTree = new FunctionTree<3, ComplexDouble>(this->getMRA());
+FunctionTree<D, ComplexDouble>* FunctionTree<D, T>::CopyTreeToComplex() {
+    auto* outTree = new FunctionTree<D, ComplexDouble>(this->getMRA());
```
gitpeterwind, line 1210: But what happens if the output FunctionTree exists already? Then its address is lost and we have a memory leak?

ilfreddy: I see your point. But at least this is a clearer layout. In the previous implementation the opposite problems was present: if you used it with a pointer which has not been allocated yet the code crashes because you attempt to delete a non existing allocation. 
Also the signature was rather involved: passing the reference to a pointer and void return argument. This is cleaner. You pass nothing (all data required is in the object) and return the object.
I guess the issue was mostly that this was designed as a utility function to help in operations with `CompFunction` and therefore *assumed* a certain state. That's why I run into problems as soon as I tried to use it in a different context.
An even better design (IMO) would be to implement this *directly* as a constructor of a `FunctionTree<D, ComplexDouble>` and pass as an argument the object(s) with the real (and possibly the imaginary) part as well. This would handle memory issues in a cleaner way.

gitpeterwind: If you feel like implementing the constructor, that would be good.
Or simply check that the tree is empty, and if it is not, give an error message that tells the user to remove the tree first.
(and it is allowed to delete a nullptr!)

ilfreddy: That's what I wanted to do in the first place 😺 
As for checking that the tree is empty, that cannot be done inside the function because the function has no concept of it. The caller should rather do that. That is more meaningful because if you return a pointer to a newly allocated object, it is rather common sense to ask yourself if the pointer you had before calling the method already contained something (and would be lost) or not.

ilfreddy:
@gitpeterwind This can go for my part now. So that I can continue with the constructor (in a second PR)

gitpeterwind (approved):
Excellent!

PR #280, opened 2026-05-04
Upgrade ft

ilfreddy:
Adds constructor of complex function tree from two real function trees, passed as input to the constructor.

The current solution is not ideal as it relies on a pre-existing function, which converted a real tree to a (real part only) complex tree. Ideally this part should be incorporated in the constructor and the operation a+ib carried node-by-node in the constructor directly.

It also provides a test file for the `FunctionTree`constructors.


tests/trees/function_tree_constructors.cpp
```diff
@@ -0,0 +1,278 @@
+#include "catch2/catch_all.hpp"
```
msnik1999: License header missing

ilfreddy:
@msnik1999 Done for my part. Can you review?

msnik1999 (approved):

PR #281, opened 2026-05-29
Modernize MRCPP

### LINKED ISSUE #77: Fix warnings appearing when compiling with clang
Fix warnings appearing when compiling with clang

Possible theme for the next hackaton

### LINKED ISSUE #191: Static code analysis (MRCpp)
compiling mrcpp yields some compiler warnings in Fedora 35. The most common issue appears to be the comparison between signed and unsigned which arises from the use of `int` as loop indices checked against `.size()` which returns the size type, `size_t`. The loop indices could probably be all switched to `size_t`, what do you think?

Additionally, running static code analysis on the repository yields unclean results; output from cppcheck 2.6 attached
[cppcheck.log](https://github.com/MRChemSoft/mrcpp/files/7791262/cppcheck.log)
.

msnik1999:
Complementary to this [MRChem PR](https://github.com/MRChemSoft/mrchem/pull/560)

Modernizing MRCPP

- [x] clean-up cmake
  - more user-accessible cmake options
  - "unify" cmake files (make different cmake files more similar)
  - remove warnings with cmake 4.x.x
  - more cmake output (progress / status)
- [x] locking current upper versioning in libraries 
  - [x] cmake
  - [x] compilers (Intel a bit unclear, as "classic" ICC compilers are no longer updated)
  - [x] Eigen3
  - [x] Catch
 - [x] Updating install documentation
 - [x] Switch to CPM


docs/install.rst
```diff
@@ -1,42 +1,60 @@
...
+C++ dependencies
+----------------
 
-The configuration and build process is managed through CMake, and a ``setup``
-script is provided for the configuration step. MRCPP's only dependency is
-Eigen3, which will be downloaded at configure time unless it is already
-available on the system. If you have a local version *not* under the system
-path, you can point to it before running ``setup``::
+- Linear algebra: `Eigen-3.4  <https://gitlab.com/libeigen/eigen>`_
+- BLAS (optional)
```
robertodr: we actually do not set any of the Eigen env-vars that would force it to link to an external BLAS, so not only it's optional, but even if it's there it's not used. Maybe remove? It's a fossil, I believe.

external/upstream/fetch_eigen3.cmake
```diff
@@ -1,20 +1,25 @@
-find_package(Eigen3 CONFIG QUIET
-  NO_CMAKE_PATH
-  NO_CMAKE_PACKAGE_REGISTRY
-  )
-
+if(EIGEN3_FIND_BEHAVIOUR STREQUAL "default" OR EIGEN3_FIND_BEHAVIOUR STREQUAL "onlylocal")
```
robertodr: same comment as in the PR on MRChem. I think these days using CPM is the better option.

robertodr (approved):
thanks @msnik1999! same comment as the other PR, it'd be preferable to use CPM instead of hand-written `fetch*.cmake` scripts

msnik1999:
> thanks @msnik1999! same comment as the other PR, it'd be preferable to use CPM instead of hand-written `fetch*.cmake` scripts

I had never heard of it before, but it looks like a very good alternative. We will look into that!

external/upstream/fetch_eigen3.cmake
```diff
@@ -1,20 +1,10 @@
...
-  set(BUILD_TESTING OFF CACHE BOOL "" FORCE)
-  FetchContent_MakeAvailable(eigen3)
-endif()
+cpm_set_find_behaviour(${EIGEN3_FIND_BEHAVIOUR})
+CPMAddPackage(
+  NAME eigen3
+  VERSION "3.4...<4.0"
+  GIT_TAG 3.4.0
+  GITLAB_REPOSITORY libeigen/eigen
+  FIND_PACKAGE_ARGUMENTS "CONFIG NO_CMAKE_PATH NO_CMAKE_PACKAGE_REGISTRY NO_CMAKE_SYSTEM_PACKAGE_REGISTRY"
```
robertodr, line 7: I think these are redundant with CPM
```suggestion
```

tests/CMakeLists.txt
```diff
@@ -1,17 +1,13 @@
...
-  FetchContent_Declare(
-    catch2
-    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-    GIT_TAG        v3.8.0
+set(CPM_USE_LOCAL_PACKAGES ON)
+CPMAddPackage(
+  NAME Catch2
+  VERSION 3.8.0
+  GITHUB_REPOSITORY catchorg/Catch2
+  FIND_PACKAGE_ARGUMENTS "CONFIG"
```
robertodr, line 6: likely redundant as well
```suggestion
```

robertodr (approved):
LGTM, it can be merged as-is, or you can address my two last comments, up to you.

Thanks again for this effort @msnik1999

external/upstream/fetch_eigen3.cmake
robertodr: Unrelated: is there a plan to switch to the latest version of Eigen? There might be quite a lot of bugfixes/perf improvements to be had!

msnik1999: Yes, but we first wanted to make sure this one works properly. When Susi tried to switch to the newer version, several pipelines failed, so we want to address that separately.

msnik1999: > I think these are redundant with CPM

I think these just prevent picking up any stray installations, e.g. from previous builds in other folders. From what I know, CPM is just a passthrough here, so CPM would not prevent any such behavior.

PR #285, opened 2026-06-19
Templatize operator trees on scalar type T; add native complex Schrod…

Valentyn113:
…inger kernel

Templatize OperatorNode, OperatorTree, CornerOperatorTree, MWOperator, ConvolutionOperator and TimeEvolutionOperator on a scalar type T (default double), so the Schrodinger semigroup can be stored as a single complex kernel (cos + i*sin) instead of a real cos/sin split.

- Behavior-preserving for real operators: the 29 existing tests are unchanged.
- TimeEvolution_CrossCorrelationCalculator emits the full complex coefficient for T = ComplexDouble (if constexpr), the real/imag part otherwise.
- New test complex_schrodinger_evolution_operator verifies the complex kernel equals the real cos + i*sin split node-for-node (30/30).
- Native complex apply is NOT included: ConvolutionCalculator's operator is still real, so applying a complex operator needs OperatorState work.


src/operators/ABGVOperator.cpp
ilfreddy, line 30: Add the following include statement:
`#include "api/mrcpp_declarations.h"`

ilfreddy: 1. The only files which need "real" changes are `OperatorNode`, `OperatorTree`, `MWNode` and `MWTree`.
2. For all other files it *should* be enough with the include statement
`#include "api/mrcpp_declarations.h"`

